PR #1, opened 2018-09-25
TSS-3077: Create salobj high level SAL Python library

r-owen:


tribeiro:
Quick suggestion, what do you think about placing the controller and the remote into modules? I think it would make the directory tree cleaner.

I mean, instead of having:

`python/salobj/controller.py` ➡️ `python/salobj/controller/controller.py`
`python/salobj/controller_command.py`  ➡️ `python/salobj/controller/controller_command.py`
`python/salobj/controller_event.py`  ➡️ `python/salobj/controller/controller_event.py`
`python/salobj/controller_telemetry.py`   ➡️ `python/salobj/controller/controller_telemetry.py`

and the same for remote.

tribeiro:
If I try to import the code for another component that's not `Test` I experience a kernel crash. Looks like it is some kind of library conflict because the error goes away if I delete `python/SALPY_Test.so`.

r-owen:
Thank you for the helpful review @tribeiro 

I removed the libraries from salobj. I should never have added them, since this package doesn't build them itself; it was a hack that backfired.

Regarding code reorganization: I aim for module names and hiearchy that makes it easy to predict where a file should be found, and putting the important classes at the top level. I'm not happy with salobj.remote.remote.Remote, for instance. But what if, instead, I add a directory "topics" and put the commands, events and telemetry modules in that. Users never create these classes directly and it moves 6 files down a level.

Unfortunately I was unable to reproduce your problem with scons. That said, I found and fixed a different problem: pytest was getting confused by a cached file path for test_utils.py and this turned out to be an unwanted side effect of a test file that has the same name as a library file, so I renamed tests/test_utils.py to tests/test_basics.py. Avoiding duplicate file names makes it easier to edit the code, so I was glad for an excuse to make that change.

tribeiro:
I think you can avoid the multilevel imports by adding import statements on the `__init__` file. 

I may be wrong but if you move the files to `python/salobj/controller/` and add 

`from .remote import *` to `python/salobj/__init__.py`, you should be able to do `import salobj.remote.Remote` or even `import salobj.Remote`.

That said, I am also ok with leaving the files where they are... Seems too much of an overhead for small gain...

r-owen:
I moved the topics one level down as I suggested and I think it makes the code look *much* nicer. It really cleans up the imports inside Remote and Controller without hiding anything (though I'm having trouble with sphinx linking to anything not at the top level; I'll talk to Jonathan Sick about that).

PR #2, opened 2018-09-30
TSS-3111: Add a base class for CSCs to salobj

r-owen:


pauljlotz:
Specific transition comment:
EnterControl: There is a definition of this but this is not used, I think because this is for the SAL Only Commands state machine, which is fine.
Note that for the SAL and Other Commands version where this is used, the transition is not from OfflineState (as in the def),

>     def do_enterControl(self, id_data):
>         """Transition from `State.OFFLINE` or `State.FAULT`
>         to `State.STANDBY`.

but from AvailableState, which is a substate of OfflineState.

pauljlotz:
General trigger comment:
A SignalEvent trigger may result in a state transition, as happens for the all the triggers in the top-level state machine. Such a trigger may, however, result in one of the following:
1) internal transition (executes a behavior without leaving a state)
2) self-transition (effectively leaves and reenters same state, executing exit and entry methods in the process).

Internal transitions are quite common, self-transitions less so. Will this framework support either or both?

pauljlotz:
ChangeEvent comment:
The only ChangeEvent in the top-level state machine is handled:

>     def fault(self):
        """Enter the fault state."""
        self.state = State.FAULT
        self.report_summary_state()

I can see how a programmer would add a ChangeEvent in the specific code called for a specific transition. Certain change events can happen any time, not just in an operation in response to a SignalTrigger, or, more often, are checked at only the highest common level in a hierarchical state machine. This is something important to think about moving forward.

pauljlotz:
General comment on state method names.
This code uses 'do' for any response to a SignalTrigger. That is nominally fine, but be aware that common state machine designs (Harel state machines as codified in the UML specification) define entry, exit, and do methods. If we want to use this distinction later, using 'do' generally would confuse this.
While I do not think it is necessary to implement purely formal entry and exit methods, the distinctions in the types of behaviors they represent are quite valuable.

r-owen:
I will remove the `enterControl` command and change `standby` to transition from FAULT to STANDBY.

With regards to "do", I think this will not be an issue for real code. I've never seen a controller use "do" as a command name. It could be a prefix, but even that is unlikely, and the result is well defined (though it has a bit of stutter). "do" is a short and clear prefix and I prefer it to "impl" or "exec" or the fully spelled-out versions of those.

I think you figured this out already, but the CSC is implemented as a subclass of BaseCsc and that subclass is responsible for handling all additional commands. Such commands may certainly change a more detailed state. In that common situation BaseCsc provides hooks (e.g. begin_<command>) in order to allow the standard commands which affect summary state to perform additional work, such as halting operations and making detailed state transitions.

tribeiro:
Sorry but I don't I understand the issue with `enterControl`. 

Why wouldn't `enterControl` be relevant to this kind of CSC? And what "kind of CSC" are we talking about? I think that as a Test case it should implement all the default states and commands. Isn't that the case? Also, isn't it after a transition from offline_available to standby (following an `enterControl` command) that a CSC is expected to broadcast the `settingVersions` event? I actually think that *this* is a missing feature on the `BaseCsc` class, maybe why `enterControl` was not being used.

python/salobj/test_utils.py
```diff
@@ -131,26 +137,38 @@ def do_setArrays(self, id_data):
...
+        """Execute the wait command.
+
+        Wait for the specified time and then acknowledge the command
+        using the specified ack code.
+        """
+        asyncio.ensure_future(self._impl_wait(id_data))
         ack = self.controller.cmd_wait.AckType()
-        ack.ack = self.salinfo.lib.SAL__CMD_INPROGRESS
+        ack.ack = self.controller.salinfo.lib.SAL__CMD_INPROGRESS
         return ack
```
tribeiro: It took me a little while to understand the fact that `do_wait` returns an `ack` when no other `do` method return anything. Wouldn't it be better to add a `return None` to all the `do` methods so it is more explicit that the method should actually return something?

python/salobj/base_csc.py
```diff
@@ -0,0 +1,336 @@
...
+        * Changes the state and reports the new value.
+        * Calls ``after_<name>``. Again, this is a no-op in the base class,
+            and is available for the subclass to override.
+        * Report the command as complete
+
+    Standard CSC commands and their associated summary state changes:
+
+    * ``start``: `State.STANDBY` to `State.DISABLED`
+    * ``enable``: `State.DISABLED` to `State.ENABLED`
+
```
tribeiro: I said that in the comments but I'll add it here again, can we take a minute to talk about the `OFFLINE` to `STANDBY` transition? Also, it would be good to add some default boiler plate to publish the `settingVersions` event. We can even talk about defining some global library to deal with the settings.

pauljlotz:
> Why wouldn't `enterControl` be relevant to this kind of CSC?

See LSE-209 2.3.1.1.

> Also, isn't it after a transition from offline_available to standby (following an enterControl command) that a CSC is expected to broadcast the settingVersions event?

No, the requirement is to publish this information "upon connecting to SAL/DDS".  LSE-209 3.2.4.2.4

tribeiro:
What about LSE-209 2.3.2.1.1?

pauljlotz:
> What about LSE-209 2.3.2.1.1?

For the SAL Only Commands version of the top-level state machine, the initial transition is to StandbyState. There is no EnterControl (or other transition) from OfflineState to StandbyState, which is unsurprising since an effect of Exit in this model is "Shut down controller."

pauljlotz:
> With regards to "do", I think this will not be an issue for real code. I've never seen a controller use "do" as a command name. It could be a prefix, but even that is unlikely, and the result is well defined (though it has a bit of stutter). "do" is a short and clear prefix and I prefer it to "impl" or "exec" or the fully spelled-out versions of those.

To clarify, I was not talking about naming methods as 'do'. This is a category of methods defined on a state, and implies certain behavioral expectations. (14.2.3.4.3 in the UML 2.5.1 specification).
Again, this is for future consideration.

tribeiro:
Oh, I see... interesting... ok, thanks for clarifying that to me!

python/salobj/test_utils.py
r-owen: I'm sorry you found it confusing, but I feel that explicit "return None" adds clutter that many Python programmers would find disturbing. I think I should clarify the documentation, instead.

As an aside, I am becoming convinced that allowing `do_<name>` to be an asynchronous function will really simplify some code. If I can manage that, then I doubt there will be any need to support returning an Ack from a `do_<name>` method.

PR #21, opened 2018-12-04
TSS-3375: Include logging in BaseCsc

r-owen:


r-owen:
@couger01 I don't understand your suggestion. That is exactly what `LogMixin` does. Based on private conversations I have pushed some changes, however: logging is now part of `Controller` (and since we no longer use it as a mixin class, it is called `Logger`). I also now log errors in command callbacks and have a test for this.

I confess I am somewhat uneasy about adding a `do_setLogLevel` method to Controller, since `do_<name>` are only relevant if callbacks are used. But I'm not sure how else to do this and still let Controller reliably have a log.

PR #25, opened 2019-01-04
TSS-3440: Make BaseCsc.implement_simulation_mode a coroutine

r-owen:


python/lsst/ts/salobj/base_csc.py
```diff
@@ -177,14 +184,21 @@ class BaseCsc(Controller):
...
 
+    In unit tests, wait for `start_task` to be done, or for the initial
+    summaryState event, before expecting the CSC to be responsive.
+
     .. _simulation_mode:
 
-    Simulation Mode:
+    **Simulation Mode**
 
     CSCs should support a simulation mode it practical; this is especially
```
couger01: Typo, "CSCs should support a simulation mode if practical; this is especially important if the CSC talks to the hardware"

PR #30, opened 2019-01-29
DM-17533: Make topic destructors more robust

r-owen:


python/lsst/ts/salobj/base.py
```diff
@@ -168,5 +168,6 @@ def makeAck(self, ack, error=0, result=""):
 
     def __del__(self):
         # use getattr in case this is called during failed construction
-        if hasattr(self, "manager"):
-            self.manager.salShutdown()
+        manager = getattr(self, "manager", None)
+        if manager:
```
tribeiro: Is it safe to do this? I remember a recommendation to do a comparison to None.

```python
if manager is not None:
```

r-owen: I think it is safe; manager is a C++ wrapped object that should never logically be False and I believe that tasks are always logically true. But I agree the code it's probably better to compare to None. I'll change all destructors accordingly.

PR #32, opened 2019-01-31
DM-17560: Add force_output argument to ControllerEvent.set_put

r-owen:


python/lsst/ts/salobj/topics/controller_event.py
```diff
@@ -58,34 +56,34 @@ def put(self, data=None, priority=1):
...
 
-        The data is put if it has never been set (`has_data` False)
-        or if this call changes any field.
+        The data is put if it has never been set (`has_data` False), or this
+        call changes the value of any field, or ``force_output`` is true.
+
+        Parameters
+        ----------
+        force_output : `bool` (optional)
+            If True the output even if not changed.
```
couger01: "If True, show the output even if not changed"?

r-owen: Agreed. I actually meant to write "If True then output even if not changed." but that might be a bit terse. I will change it to "If True then output the event, even if not changed".

PR #33, opened 2019-02-14
DM-17913: Add a function to ts_salobj that enables a CSC from any state

r-owen:
Ref: [https://jira.lsstcorp.org/browse/DM-17913](url)

### DIFF
--- a/bin.src/purge_topics.py
+++ b/bin.src/purge_topics.py
@@ -1,0 +1,196 @@
+#!/usr/bin/env python
+# This file is part of ts_salobj.
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+import argparse
+import asyncio
+import importlib
+import concurrent.futures
+import sys
+import time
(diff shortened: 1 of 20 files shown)


bin.src/purge_topics.py
```diff
@@ -0,0 +1,120 @@
...
+        Index of SAL component.
+    commands : `list` of `str` (optional)
+        Names of comamnds to purge; if None then purge all commands
+    sleep : `float`
+        Amount of time to sleep after purging each command,
+        before closing the SAL object (sec).
+    """
+    lib = importlib.import_module(f"SALPY_{component}")
+    if commands is None and events is None and telemetry is None:
+        salinfo = salobj.SalInfo(lib, index)
```
tribeiro: I think `salinfo` has to be define before the `if` statement, otherwise the script will fail if any one of the commands, events or telemetry is passed.

bin.src/purge_topics.py
```diff
@@ -0,0 +1,120 @@
...
+        telemetry = ()
+
+    for command_name in commands:
+        salinfo = salobj.SalInfo(lib, index)
+        topic_name = salinfo.name + "_command_" + command_name
+        try:
+            salinfo.manager.salProcessor(topic_name)
+        except Exception as e:
+            print(f"Could not purge command {command_name}: salProcessor raised {e}")
+            continue
```
tribeiro: With the way it "works" now, once it fails for one topic, it will fail for the rest. So, what I would recommend doing here is stoping and, stretch goal, print a list of the remaining topics to clear.

python/lsst/ts/salobj/topics/controller_command.py
```diff
@@ -279,6 +279,4 @@ def _setup(self):
         self._DataType = getattr(self.salinfo.lib, self.salinfo.name + "_command_" + self.name + "C")
 
         topic_name = self.salinfo.name + "_command_" + self.name
-        retcode = self.salinfo.manager.salProcessor(topic_name)
-        if retcode != self.salinfo.lib.SAL__OK:
-            raise RuntimeError(f"salProcessor({topic_name}) failed with return code {retcode}")
+        self.salinfo.manager.salProcessor(topic_name)
```
tribeiro: Do you think you can catch the exception and then trow it again but printing which topic failed? the current message obscures the failed topic so we need to go look at the log file to figure out. Either this of do it inside SAL.

r-owen: Yes. It would be better if SAL included the information itself, but that may be awhile before that happens so I have added it here.

PR #37, opened 2019-03-20
DM-18344: Add support for configurable CSCs

r-owen:


python/lsst/ts/salobj/base_csc.py
```diff
@@ -33,7 +33,7 @@
 
 
 class State(enum.IntEnum):
```
tribeiro: Please, add a comment as to why not using the enumeration values from SALPY.

python/lsst/ts/salobj/configurable_csc.py
```diff
@@ -0,0 +1,151 @@
...
+      by specified ``schema_path``.
+      This includes setting default values from the schema and validating
+      the result again (in case the default values are invalid).
+    * The validated configuration is converted to a struct-like object
+      using `types.SimpleNamespace`.
+    * The configuration is passed to the ``configure`` method,
+      which subclasses must override. Note that ``configure`` is called just
+      before summary state changes from `State.STANDBY` to `State.DISABLED`.
+    """
+    def __init__(self, sallib, index, config_dir, schema_path,
```
tribeiro: CSCs should publish a `settingVersions` event with the `recommendedSettingsVersion` and `recommendedSettingsLabels`. 

I think that a Configurable CSC has enough information to publish this event so I think it would be great to do that by default. One caveat is that it must be re-published anytime the CSC goes into standby.

The `recommendedSettingsLabels` contains all available settings and `recommendedSettingsVersion` the version of the configuration. 

This brings the question as to how we manage the version of a configuration. I think that right now we could use the `__fingerprint__` in `version.py`.

r-owen: Where are these documented? I see them in SALGenerics.xml but they have no description. I'm especially puzzled by `recommendedSettingsVersion` since the only thing the `start` command takes is a single string which is the name of the config file. It seems to me that users need a list of the names that can be provided to `start`. From this I conclude:

• `recommendedSettingsLabels` should be a list of files in the config directory
• If the config directory contains multiple versions of a given config then they need to be differentiated by file name (easily done)

Another worry is that `recommendedSettingsLabels` is only 256 chars long. So it could easily be too short to list all available configuration files, especially if there is a hierarchy.

tribeiro: That is a very good point. I'm looking for the documentation regarding this and will let you know about my findings. What I know is from meetings with SE team that led the effort on developing that topic.

r-owen: Thank you. I'm glad you caught this omission and I clearly need to add support for it. I'm just not sure how to go about it.

tribeiro: We can probably leave it as it is for now and deal with that later once we have a better understanding of the solution.

r-owen: I just talked to @CContaxis about this, since he had to deal with it for M1M3. To summarize what he does (I've copied him on this so he can correct any errors):
- `recommendedSettingsLabels` contains a list of the recommended settings labels. It does not include other versions that are not recommended.
- Old versions are always retained.
- M1M3 uses an alias so the real files always have a version name and the alias probably does not. This has the advantage that you can always use the same string to mean the same basic configuration, but it has the disadvantage that you cannot tell which file was actually used. I think if we go that route then we must use the `recommendedSettingsVersion` field to identify the git version or something. Still...I prefer explicit to implicit.

My own inclination is to make the entries in `recommendedSettingsLabels` include include version information (and not use aliases). I am less ure what to put in `recommendedSettingsVersion` but I also doubt it matters very much.

python/lsst/ts/salobj/configurable_csc.py
```diff
@@ -0,0 +1,253 @@
...
+    """Base class for a configurable Commandable SAL Component (CSC)
+
+    Parameters
+    ----------
+    sallib : ``module``
+        salpy component library generatedby SAL
+    index : `int` or `None`
+        SAL component index, or 0 or None if the component is not indexed.
+    config_dir : `str`
+        Path to configuration files.
```
couger01: Should a config_directory be able to be changed, or is it set once and then thats it.

python/lsst/ts/salobj/configurable_csc.py
```diff
@@ -0,0 +1,253 @@
...
+        id_data : `CommandIdData`
+            Command ID and data
+        """
+        config_name = id_data.data.settingsToApply
+        if config_name:
+            config_file_name = self.config_label_dict.get(config_name, config_name)
+            config_file_path = self.config_dir / config_file_name
+            if not config_file_path.is_file():
+                raise base.ExpectedError(f"Cannot find config file {config_file_name} in {self.config_dir}")
+            with open(config_file_path, "r") as f:
```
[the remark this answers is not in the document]
r-owen: One configuration file. This meets the requirement that a full configuration can be specified by a single label. If you want to use that one file to specify other files then you can do that, but `ConfigurableCsc` will not support that -- you would have to override at least one method to make that work.

python/lsst/ts/salobj/configurable_csc.py
r-owen: It is set when the CSC is constructed. Normally the `config_dir` constructor argument will be `None` and `ConfigurableCsc` will work out where to find the correct directory (see the last comment on this ticket for more information). But that code was not in place when you commented on this pull request.

python/lsst/ts/salobj/configurable_csc.py
```diff
@@ -0,0 +1,376 @@
...
+            as a struct-like object.
+
+        Notes
+        -----
+        Called when running the ``start`` command, just before changing
+        summary state from `State.STANDBY` to `State.DISABLED`.
+        """
+        raise NotImplementedError("Subclasses must override")
+
+    @classmethod
```
tribeiro: It seems to me that this could be a `staticmethod` instead.

r-owen: Yes, I thought about that. I felt it a bit safer to leave it a classmethod in case the code wanted to somehow auto-generate it based on class information. But I freely admit that is unlikely and will happily change it if you wish.

tribeiro: Yes, I thought about that as well. I think a `staticmethod` will pass a better message, which is "don't try to be clever about this". :)

r-owen: OK. I have changed it (as new commits)

PR #41, opened 2019-04-17
DM-18860 Make start_/end_ and configure methods async

r-owen:


python/lsst/ts/salobj/base_csc.py
```diff
@@ -248,39 +249,39 @@ def add_kwargs_from_args(cls, args, kwargs):
         """
         pass
 
-    def do_disable(self, id_data):
+    async def do_disable(self, id_data):
         """Transition to from `State.ENABLED` to `State.DISABLED`.
```
tribeiro: "Transition to from" ➡️ "Transition from"

r-owen: Good catch; I fixed that and another instance

PR #42, opened 2019-04-20
DM-19385: use dds and sal-generated IDL files instead of SALPY libraries

r-owen:


python/lsst/ts/salobj/domain.py
```diff
@@ -0,0 +1,250 @@
...
+                host = int(host)
+            except ValueError:
+                raise ValueError(f"Could not parse $LSST_DDS_IP={host} as an integer")
+        self.host = host
+        """Value for the private_host field of output samples."""
+
+        self.origin = os.getpid()
+        """Value for the private_origin field of output samples."""
+
+        idl_dir = os.environ.get("TS_IDL_DIR")
```
tribeiro: @r-owen , do you think you can change the strategy to get the idl files? As from our previous conversation, I’m not a big fan of using environment variables… I have two proposed suggestions, in order of preference:

1 - Remove ts_idl and add an “idl” directory at the top level of salobj where the idl and qos files are stored. If this is the case, "Test" idl files could even be part of the package, along with the qos files. 
2 - Keep ts_idl but add a python package that defines the paths, then import the path from the idl package. Something like:

```
from lsst.ts.idl import idl_dir
```

and idl_dir could be defined like:

```
idl_dir =  pathlib.Path(__file__).resolve().parents[4]
```

or something like that.

python/lsst/ts/salobj/domain.py
```diff
@@ -0,0 +1,250 @@
...
+            self.participant = None
+
+    def tai_utc(self, utc):
+        """Return TAI-UTC (seconds) given UTC in unix seconds.
+
+        Warnings
+        --------
+        This is a temporary hack.
+        """
+        return 37
```
tribeiro: Maybe add a `#FIXME` tag?

python/lsst/ts/salobj/domain.py
r-owen: I think all languages need access to the QoS file, which means putting it in ts_salobj or using a python path is not a good solution for that. I should check to see what other languages are doing and consider using that instead of putting it in ts_idl.

I am uneasy about putting the IDL files in ts_salobj, because they change on a different time scale than ts_salobj itself. This may make XML and salobj updates a bit complicated.

Personally I think environment variables are better than relying on Python paths for getting this kind of data. In theory other languages might also want to get to these IDL files (though I think it unlikely), or we might want to put other kinds of SAL libraries in this package (which I think would be great, but then the name isn't ideal). Neither use case can use the python `module.__file__` trick.

python/lsst/ts/salobj/domain.py
r-owen: Good suggestion. I filed DM-19791 and added a TODO reference to it.

python/lsst/ts/salobj/domain.py
r-owen: I asked @dmills-lsst about the qos file `DDS_DefaultQoS_All.xml` and he said that he plans to transition all CSCs to use that file, and plans to use environment variable `LSST_DDS_QOS` to point to it. I am a bit worried about where the file should live for ts_salobj but I should probably go along with this plan and modify ts_salobj to use this. I'm not sure how ts_salobj can safely find the shared common QoS file except by using an environment variable.

python/lsst/ts/salobj/base.py
```diff
@@ -126,90 +111,16 @@ def index_impl():
...
-                raise ValueError(f"len(result) > MAX_RESULT_LEN={MAX_RESULT_LEN}; result={result}")
-        data.result = result
-        return data
+def tai_minus_utc(utc):
+    """Return TAI unix seconds, given UTC in unix seconds.
+
+    This function is only intended for current time;
+    it makes no attempt to be correct for historical dates.
+    """
+    return utc + 37
```
tribeiro: Should this be also part of DM-19791? Maybe it would be a good idea to add a `#TODO` or a `#FIXME` note.

r-owen: We only need one of the two (and this one is presently broken). I will combine them into one free function named `tai_from_utc` since that is how it is always used.

python/lsst/ts/salobj/domain.py
r-owen: I have updated ts_idl to include the following:
* get_idl_dir() a function that returns the root dir of the package
* get_qos_path() a function that returns the path to the QoS file; for now it points to a copy in ts_idl but that could change if @dmills-lsst feels strongly enough about it.
* lsst.ts.idl.enums: a set of enum files for various CSCs. For now I have enums for ATDome, ATMCS, Script and ScriptQueue. I will add more as needed. They are manually generated for now and I suspect we'll just stick to that, but if we can generate them automatically (preferably without changing the names for the classes and enum constants) then that would be even better.

python/lsst/ts/salobj/remote.py
```diff
@@ -34,8 +37,15 @@ class Remote:
```
tribeiro: There is a typo here in 

"If a SAL **conponent** listens to or commands other SAL components"

r-owen: Fixed. I also looked for other instances of the same misspelling and did not find any.

PR #52, opened 2019-06-28
DM-20327: Build/Deployment framework design

tribeiro:
Adds a Jenkinsfile and Dockerfile to be used to drive CI test.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,53 @@
+ARG base_image_tag
+
+FROM lsstts/salobj:$base_image_tag
+
+# Arguments for package versions
+ARG config_ocs_v="develop"
+ARG sal_v="develop"
+ARG salobj_v="develop"
+ARG xml_v="develop"
+ARG idl_v="develop"
+
+WORKDIR /home/saluser/repos/ts_config_ocs
+RUN /home/saluser/.checkout_repo.sh ${config_ocs_v}
+
+WORKDIR /home/saluser/repos/ts_sal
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+RUN /home/saluser/.checkout_repo.sh ${xml_v}
+
+WORKDIR /home/saluser/repos/ts_idl
+RUN /home/saluser/.checkout_repo.sh ${idl_v}
+
+WORKDIR /home/saluser/repos/ts_sal
+RUN source /opt/lsst/software/stack/loadLSST.bash && setup lsst_distrib && \
+    source /home/saluser/repos/ts_sal/setup.env && \
+    setup ts_sal -t current && \
+    scons
```
r-owen: Why do you run `/opt/lsst/software/stack/loadLSST.bash` each time instead of once before doing the rest? Is your environment completely reset for each RUN command?

r-owen: I'm curious: I see that the `setup xxx -t current` commands work because of the `setup.env` files (rather than using the `eups declare` commands I expected), but where do the `setup.env` files come from? Also, how do you make sure the `python/lsst/ts/.../version.py` file is created in each repo?

tribeiro: yes, unfortunately, you do have to source and setup everything each time your make a `RUN` statement.  

This Dockerfile starts from an already built salobj container. So, in principle, all the packages are already installed and ready to go. The goal is to checkout each one of the dependencies and rerun the build.

PR #61, opened 2019-09-23
DM-21358: only allow setSimulationMode in the STANDBY state

r-owen:


doc/salobj_cscs.rst
```diff
@@ -226,6 +226,22 @@ Note that this method is called during construction of the CSC.
...
+
+* If the current state is DISABLED or ENABLED state and not already connected, then make the connection.
+  If you support simulation mode then read that to determine if this is a real or a simulated connection.
+* If the current state is something else then disconnect.
+
+Examples include the following (both of which have a simulation mode):
+
+* ts_ATDome talks to a TCP/IP controller
+* ts_FiberSpectrograph controls fiber spectrographs over USB.
+
```
couger01: Food for thought. I've made and break the connection in the tunable_laser by using the hooks in the state transitions. This allows the configuration settings in the start command to change the port of the laser.

r-owen: I strongly prefer to use `report_summary_state` instead of overriding `begin_x` and `end_x` methods, because it puts all the related code in one place.

Note that this is orthogonal to getting the connection information from the configuration.

couger01: Fair point and true about the connection information. But I do think the hook way is clearer to me about what is happening in each state transition. You're good to merge when you want.

PR #63, opened 2019-09-26
DM-21304 add async BaseCsc.handle_summary_state

r-owen:


r-owen:
The integration check is probably just needing to temporarily add ticket branches to ts_sal, ts_idl and ts_xml. A side effect of the delay in releasing SAL 4. But we'll see. I have added those branches and done a `git commit --amend` to trigger a new Jenkins run.

PR #65, opened 2019-10-31
DM-22037: have Controller wait for remotes.

r-owen:


python/lsst/ts/salobj/base_script.py
```diff
@@ -87,34 +87,28 @@ def __init__(self, index, descr):
...
-        """A task that is set to None, or an exception if cleanup fails,
-        when the task is done.
-        """
         self._is_exiting = False
         self.evt_description.set(
             classname=type(self).__name__,
             description=str(descr),
         )
         self._heartbeat_task = asyncio.ensure_future(self._heartbeat_loop())
+        # Delay (sec) to allow sending the final state and aknowleding
```
tribeiro: typo in `aknowleding`

python/lsst/ts/salobj/domain.py
```diff
@@ -139,15 +140,19 @@ def __init__(self):
         # set of SalInfo
         self._salinfo_set = weakref.WeakSet()
 
-        host = os.environ.get("LSST_DDS_IP")
-        if host is None:
+        host_name = os.environ.get("LSST_DDS_IP")
```
tribeiro: What happens if LSST_DDS_IP is not set?

python/lsst/ts/salobj/domain.py
```diff
@@ -139,15 +140,19 @@ def __init__(self):
...
+        if host_name is None:
             host = random.randint(1, MAX_RANDOM_HOST)
         else:
             try:
-                host = int(ipaddress.IPv4Address(host))
+                unsigned_host = int(ipaddress.IPv4Address(host_name))
             except ipaddress.AddressValueError as e:
-                raise ValueError(f"Could not parse $LSST_DDS_IP={host} "
+                raise ValueError(f"Could not parse $LSST_DDS_IP={host_name} "
```
tribeiro: I guess the answer to my question above is here... I'm not entirely sure the best thing to do is to raise an exception. Seems like too harsh to prevent the communication from happening because you can't parse `LSST_DDS_IP`... But I can see it is useful to be able to uniquely identify the source.

tests/test_basics.py
```diff
@@ -124,15 +124,25 @@ def test_set_random_lsst_dds_domain(self):
...
-            os.environ["LSST_DDS_IP"] = bad_ip
-            with self.assertRaises(ValueError):
-                salobj.Domain()
-
-        # a value from the ipaddress documentation
-        os.environ["LSST_DDS_IP"] = "192.168.0.1"
-        async with salobj.Domain() as domain:
-            self.assertEqual(domain.host, 3232235521)
-            self.assertEqual(domain.origin, os.getpid())
+            with self.subTest(bad_ip=bad_ip):
```
tribeiro: and there's even a test for when you can't parse it... :)

python/lsst/ts/salobj/domain.py
r-owen: A random number is used

python/lsst/ts/salobj/domain.py
r-owen: This error is raised if LSST_DDS_IP is set to a value that cannot be parsed as a dotted address. I think that's reasonable because it means the user intended to provide a useful value but it cannot be used. LSST_DDS_IP is not set to anything then the None case a bit higher up is used and a random value is chosen.

PR #66, opened 2019-11-19
DM-22269: add s3 file upload/download support to salobj

r-owen:


python/lsst/ts/salobj/async_s3_bucket.py
```diff
@@ -0,0 +1,130 @@
...
+import botocore
+
+
+class AsyncS3Bucket:
+    """Asynchronous interface to Amazon Web Services s3 buckets.
+
+    Parameters
+    ----------
+    name : `str`
+        Name of bucket.
```
parejkoj: Could you be a little more specific about what this "name" is? Is it a URI or URL? Or some internal S3 nomenclature?

python/lsst/ts/salobj/async_s3_bucket.py
```diff
@@ -0,0 +1,130 @@
...
+        """Upload a file-like object to the bucket.
+
+        Parameters
+        ----------
+        fileobj : file-like object
+            File-like object that can be read as *binary* data.
+        key : `str`
+            Name to use for the file in the bucket.
+        callback : callable (optional)
+            Function to call with updates while writing. The function receives
```
parejkoj: It might be worth mentioning that this is passed on through to the s3 `upload_fileobj` method, if someone is curious how it behaves in detail (since you have no control over that).

python/lsst/ts/salobj/async_s3_bucket.py
```diff
@@ -0,0 +1,130 @@
...
+            one argument: the number of bytes written. If the transfer is
+            successful then it will always be called at least once,
+            and the sum of the number of bytes for all calls will equal
+            the size of the file.
+
+        Notes
+        -----
+        To create a file-like object from an astropy.io.fits object:
+            fileobj = io.BytesIO()
```
parejkoj: Did you forget to put a FITS object in here? I forget, would it be `io.BytesIO(hdulist)`?

python/lsst/ts/salobj/async_s3_bucket.py
```diff
@@ -0,0 +1,130 @@
...
+            hdulist.writeto(fileobj)
+        """
+        loop = asyncio.get_event_loop()
+        await loop.run_in_executor(None, self._sync_upload, fileobj, key, callback)
+
+    async def download(self, key, callback=None):
+        """Download a file-like object from the bucket.
+
+
```
parejkoj: extra blank line

python/lsst/ts/salobj/async_s3_bucket.py
```diff
@@ -0,0 +1,130 @@
...
+        key : `str`
+            Name of the potential file in the bucket.
+        """
+        loop = asyncio.get_event_loop()
+        try:
+            await loop.run_in_executor(None, self._sync_size, key)
+        except botocore.exceptions.ClientError as e:
+            if e.response['Error']['Code'] == '404':
+                return False
+            # Some other error.
```
parejkoj: I don't understand the point of this comment? Did you mean "Raise other errors back to the caller."?

tests/test_async_s3_bucket.py
```diff
@@ -0,0 +1,119 @@
...
+        self.mock = moto.mock_s3()
+        self.mock.start()
+
+        # Set s3 authentication environment variables to bogus values
+        # to avoid any danger of writing to a real s3 server.
+        for key in ("AWS_ACCESS_KEY_ID",
+                    "AWS_SECRET_ACCESS_KEY",
+                    "AWS_SECURITY_TOKEN",
+                    "AWS_SESSION_TOKEN"):
+            os.environ[key] = "testing"
```
parejkoj: Is it worth saving and resetting these as part of `tearDown`? Or does `os.environ` only live as long as this python instance?

tests/test_async_s3_bucket.py
```diff
@@ -0,0 +1,119 @@
...
+
+        # Make an s3 bucket so tests can use it.
+        self.bucket_name = "async_bucket_test"
+        conn = boto3.resource('s3')
+        conn.create_bucket(Bucket=self.bucket_name)
```
parejkoj: Do you need to destroy this bucket in `tearDown`, or does stopping the mock handle that automatically?

tests/test_async_s3_bucket.py
```diff
@@ -0,0 +1,119 @@
...
+        roundtrip_fileobj = await bucket.download(key=key)
+        roundtrip_data = roundtrip_fileobj.getbuffer()
+        self.assertEqual(file_data, roundtrip_data)
+
+    async def test_exists(self):
+        file_data = b"Data for the test case"
+        key = "test_file"
+        bucket = salobj.AsyncS3Bucket(self.bucket_name)
+
+        should_be_false = await bucket.exists(key="no_such_file")
```
parejkoj: 👍 on the name here.

tests/test_async_s3_bucket.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class AsyncS3BucketTest(asynctest.TestCase):
+    def setUp(self):
+        # Use moto.mock_s3 in "raw mode" because:
+        # * @moto.mock_s3() does not work on async test methods.
+        # * Even if it did, or if I used moto.mock_s3() as a context manager,
+        #   I'd have to make the s3 bucket inside the test code.
+        self.mock = moto.mock_s3()
+        self.mock.start()
```
parejkoj: You have the same `file_data`, `key`, and `bucket` in each test: is it worth defining those in `setUp` and using the `self` versions in the tests?

python/lsst/ts/salobj/async_s3_bucket.py
r-owen: Good suggestion. I added a link to the relevant Amazon Web Services documentation.

python/lsst/ts/salobj/async_s3_bucket.py
r-owen: No. The full comment is:
```
        To create a file-like object from an astropy.io.fits object:
            fileobj = io.BytesIO()
            hdulist.writeto(fileobj)
```
But I'll amend the first sentence to:
```
To create a file-like object from an `astropy.io.fits` object named ``hdulist``
```

tests/test_async_s3_bucket.py
r-owen: Changing os.environ only affects the Python subprocess.

tests/test_async_s3_bucket.py
r-owen: The bucket lives in the mock, so tearing down the mock should clean up all resources.

python/lsst/ts/salobj/async_s3_bucket.py
r-owen: The only error that means "does not exist" is the one tested above; any other error is unexpected and so re-raised. I tried rewording the comment for clarity:
```
        except botocore.exceptions.ClientError as e:
            if e.response['Error']['Code'] == '404':
                return False
            # Something went wrong; report it.
            raise
```

python/lsst/ts/salobj/async_s3_bucket.py
r-owen: Also:
* I added instructions for for converting a downloaded file-like object to an HDUList.
* Modified the download code to call `seek(0)` on the file-like object before returning it.
* Modified the round-trip test to use `read` instead of `getbuffer` in order to test that the downloaded file-like object has been rewound to the beginning.

python/lsst/ts/salobj/async_s3_bucket.py
r-owen: OK. I have added it to the Notes for upload and download. I will say that the boto3 documentation is poor, and in this case it says almost nothing about how these callbacks are handled. I had to work out the details that I mention in the doc string by myself, and hope that they will not change in the future.

PR #67, opened 2019-11-22
DM-22293: add topic metadata to SalInfo

r-owen:


python/lsst/ts/salobj/idl_metadata.py
```diff
@@ -0,0 +1,232 @@
...
+                if not topic_match:
+                    continue
+
+                sal_topic_name = topic_match.group("sal_topic_name")
+                if sal_topic_name in topic_info:
+                    raise RuntimeError(f"Already parsed data for topic {sal_topic_name}")
+                version_hash = topic_match.group("version_hash")
+                topic_description = topic_match.group("description")
+
+# FIX ME: Parse the description metadata
```
afausti: I think you can remove this comment

r-owen: Good catch. Fixed.

PR #71, opened 2019-12-16
DM-22627: Rename Test and Script duration fields to timespan

r-owen:


r-owen:
Cancelling this pull request for now because we may decide we don't need the rename.

PR #72, opened 2019-12-19
DM-22714: fix a warning

r-owen:


r-owen:
I strongly recommend using `asynctest` for your unit tests. It makes the tests much easier to read and produces better output if they fail.

As far as the use of an external Harness class: it is a bit less clear-cut, but I lean towards replacing it with a `@contextlib.asynccontextmanager` test class method `make_csc`. Advantages include:
* It eliminates an external class and tends to save a bit of code.
* You can use self.foo everywhere instead of harness.foo

The main disadvantage is that `make_csc` should set attributes of self (such as `csc` and `remote`) which is a bit surprising (unittest tests normally do not set attributes of `self`). But I think it is justified in this case.

PR #77, opened 2020-01-13
DM-18126: Add setGroupId command to BaseScript

r-owen:


doc/revision_history.rst
```diff
@@ -6,6 +6,18 @@
 Revision History
 ################
 
+v5.4.0
+======
+
+Major changes:
+
+* Add support for the new ``setGroupId`` ``Script``  command and ``groupId`` Script event to `BaseScript`:
```
timj: Should the second `Script` also have double backticks around it?

python/lsst/ts/salobj/base_script.py
```diff
@@ -227,6 +230,12 @@ def checkpoints(self):
         """
         return self.evt_checkpoints.data
 
+    @property
+    def group_id(self):
+        """Get the group ID, or "" if not set.
```
timj: It seems that most properties do document the return type but this one is missing.

python/lsst/ts/salobj/base_script.py
```diff
@@ -581,6 +617,21 @@ def do_setCheckpoints(self, data):
...
+
+        Raises
+        ------
+        RuntimeError
+            If there is no group ID.
+        """
+        if not self.group_id:
+            raise RuntimeError("No group ID")
+        self._sub_group_id += 1
+        return f"{self.group_id}-{self._sub_group_id}"
```
timj: @ktlim are you okay with this form or would you prefer the sub group to be prepended? It's not relevant for AP in general since standard visits and alternate standard visits will not be changing the group ID at all, but it might be relevant for when AP tries to check a group ID (unless it filters on survey name first). OCPS and butler don't care since they treat the group ID string as being opaque.

tests/test_base_script.py
```diff
@@ -235,46 +250,73 @@ def test_non_configurable_script_get_schema(self):
 
             self.assertFalse(script.done_task.done())
 
+    async def test_next_supplemented_group_id(self):
+        async with salobj.TestScript(index=self.index) as script:
+            await self.configure_and_check(script)
+            group_id = script.group_id
+            for i in range(5):
+                desired_supplemented_id = group_id + "-" + str(i+1)
```
timj: Consider:
```python
desired_supplemented_id = f"{group_id}-{i+1}"
```

doc/revision_history.rst
r-owen: Yes, and I ended up moving groupId to the Script state event, so I need to fix this text.

python/lsst/ts/salobj/base_script.py
r-owen: I only document the return type when it is not obvious, but perhaps this one fits that category. I'll add something.

tests/test_base_script.py
r-owen: Will do. I was on the fence but I agree.

python/lsst/ts/salobj/base_script.py
ktlim: I'm OK with this.  I think Prompt Processing will need to filter on survey name, and sub-group-ids will only be used, if at all, by code that knows that they're there.

r-owen: I personally prefer a suffix in order to have group IDs sort, at least approximately.

They will not sort if there are more than 10 unless we format the sub-group ID (an integer) with leading zeros. If you want that then please tell me how many digits you want. 6 digits would handle even the most pathalogical script. I fear 2 might not be enough. I suspect anywhere in 3-5 would be fine.

r-owen: @ktlim also said privately that not having leading zeros is fine.

PR #79, opened 2020-01-27
DM-23186: add pre-commit hook

r-owen:


r-owen:
Might you suggest a better wording for the README? I want to use the same boilerplate in all my packages.

README.rst
```diff
@@ -2,7 +2,7 @@ Object-oriented interface to `Service Abstraction Layer <https://docushare.lsstc
 
 `Documentation <https://ts-salobj.lsst.io>`_
 
-The package is compatible with setuptools, as well as the Vera Rubin LSST DM's ``eups`` package management system and ``scons`` build system.
+The package is compatible with setuptools, as well as the Vera C. Rubin LSST DM's ``eups`` package management system and ``scons`` build system.
```
timj: How about:
```markdown
The package is compatible with setuptools, as well as the [``eups`` package management system](https://github.com/RobertLuptonTheGood/eups) and ``scons`` build system.
```
?

timj: (except you are using rST and not markdown...)

r-owen:
By the way: I thought pretty hard about commit hooks and I'm still a bit uneasy. There doesn't seem to be a foolproof way to automatically enable them for everyone who clones the package. I chose to this solution as being simple and not requiring the pre-commit Python package (which as far as I can see doesn't really gain us much for such a simple hook). If you have any thoughts on this I would welcome them.

timj:
I think it's best to try things out and see what happens. It's going to be interesting to see if people set things up properly. At least you should be able to spot things in the pull requests.

README.rst
r-owen: Sounds good, though I will make only "eups" the link.

PR #87, opened 2020-02-24
DM-23368: Set master priority=0 for scripts to speed up startup

r-owen:


python/lsst/ts/salobj/base_script.py
```diff
@@ -85,12 +88,17 @@ class BaseScript(salobj.Controller, abc.ABC):
     """
 
     def __init__(self, index, descr):
+        # Speed up script loading time and avoid expensive system alignments
+        # by making sure scripts never become master.
+        # This must be done before the `DomainParticipant` is created.
+        os.environ[base.MASTER_PRIORITY_ENV_VAR] = "0"
```
tribeiro: What happens if script queue stops and we start it again on the same session? will it now start with `MASTER_PRIORITY` of 0 instead of what the default is?

r-owen: Excellent question. I also worried about that. Fortunately:
* The script queue starts each script in its own subprocess, so the script queue should not be affected by this change.
* Even if you did run the script queue with master priority =0 it should not cause problems. What is important for operations is that *some* node on the system have a higher master priority than 0. It doesn't have to be the script queue.

My expectation is that we will set master priority as follows:
* Launch one or a few SAL components with a high master priority. Preferably these will be things we launch early on. Dave suggested the SAL kafka producers.
* Run most SAL components with the default value. These will rarely become master (only if none of the first category are running, which can happen in unit tests and small test systems).
* Run scripts with 0. These will never become master.

I will add a test to make sure that launching a script in a subprocess does not leak the value back to the main process.

PR #88, opened 2020-02-25
DM-23634: Add BaseConfigTestCase for testing config files in ts_config_x packages

r-owen:


wvreeven:
In principles the changes look good. I am a little surprised about replacing the call to joinpath with / characters. I know we run on Linux but being platform independent generally is much better.

r-owen:
You are right to be skeptical, but `pathlib` uses `/` as the standard path join operator, even on Windows. See https://docs.python.org/3.7/library/pathlib.html#operators

PR #90, opened 2020-03-05
DM-23771: add CscCommander

r-owen:


python/lsst/ts/salobj/csc_commander.py
```diff
@@ -0,0 +1,467 @@
+# This file is part of ts_hexrotcomm.
```
couger01: Technically should be changed to ts_salobj.

r-owen:
Thank you for catching that. I made the change. I also tweaked the arguments to CscCommander a bit:
* Made index optional, to better serve non-indexed SAL components.
* Added arguments to exclude certain topics and allow the user to tune which fields are excluded. I'm not sure how useful either will be, but I'd prefer to offer the flexibility.

PR #92, opened 2020-03-09
Implement conda packaging for salobj.

tribeiro:
Conda build is running unit tests but they are skipping SALPY tests. Need to figure out how to capture that dependency in the future so the unit tests add full coverage.

Change unittests so they are also compatible with conda packaging as well as eups/setup.


r-owen:
Note: develop had moved on, and I tried using github's button to rebase, but it did the wrong thing. So I rebased locally and made a new pull request. My apologies. At least now I know to avoid that button.

PR #93, opened 2020-03-10
DM-23757: Implement conda packaging for salobj. Conda build is running unit tes…

r-owen:
…ts but they are skipping SALPY tests. Need to figure out how to capture that dependency in the future so the unit tests add full coverage.

Change unittests so they are also compatible with conda packaging as well as eups/setup.


conda/meta.yaml
```diff
@@ -1,24 +1,49 @@
...
+    - pyyaml
+    - boto3
+    - moto
+    - ts-dds
+    - ts-idl
+  source_files:
+    - python
+    - bin
+    - tests
+    - schema
```
r-owen: Is it worth adding docs to this list?

setup.py
```diff
@@ -1,26 +1,35 @@
-from setuptools import setup, find_namespace_packages
+import os
+import sys
+import pathlib
+import setuptools
 
 install_requires = ["numpy"]
-tests_require = ["pytest", "pytest-cov", "pytest-flake8"]
+tests_require = ["pytest", "pytest-cov", "pytest-flake8", "dds"]
```
r-owen: Surely `tests_require` is extra requirements needed to run unit tests (as opposed to just using the package). But "dds" is needed to *use* the package, so shouldn't it be listed somewhere else, such as `install_requires`? Also "astropy" seems to be missing and it is in the same category as "dds".

setup.py
```diff
@@ -1,26 +1,35 @@
...
 dev_requires = install_requires + tests_require + ["documenteer[pipelines]"]
 scm_version_template = """# Generated by setuptools_scm
 __all__ = ["__version__"]
 
 __version__ = "{version}"
 """
+tools_path = pathlib.PurePosixPath(setuptools.__path__[0])
```
r-owen: I suggest using `Path` instead of `PurePosixPath`, or is there some subtlety of pathlib that I'm missing? I use `Path` everywhere and it works fine.

r-owen:
Please update revision_history.rst to describe this change. Or I will if you prefer.

Please also do this for other packages in DM-23757 that have a revision history

r-owen:
I updated the revision history.

conda/meta.yaml
tribeiro: this list in particular is what should be added for conda build test phase... So I don't think adding docs would do anything.

setup.py
tribeiro: good question... I honestly don't know... I can certainly try to replace that by `Path` and see what happens.

r-owen: I would appreciate it if you would do that.

setup.cfg
```diff
@@ -6,11 +6,13 @@ max-line-length = 110
 max-doc-length = 79
 ignore = E133, E203, E226, E228, N802, N803, N806, N812, N813, N815, N816, W503
 exclude =
-  python/lsst/ts/salobj/ddsutil.py,
```
tribeiro: Is this exclusion correct? I wonder if it might be a merge problem. Can you please verify?

r-owen: The exclusion was correct at the time because ddsutil.py was from ADLink and did not pass flake8. We switched to the version of ddsutil.py in OpenSplice (once it was updated to fix a serious segfault) and got rid of the version in salobj. Thus there is no longer any need to exclude it from flake8 testing. it's gone!

tribeiro: good! I was afraid I missed it when I did a rebase. Please, go ahead and merge this.

PR #97, opened 2020-04-13
DM-23670: output the settingsApplied and softwareVersions events

r-owen:
and also update AsyncS3Bucket to use revised bucket and key names


r-owen:
Thank you. I'll see what I can do to clean up the documentation according to your excellent suggestions. Note that we have no update documentation on the LFA so I may not be able do anything about that one.

r-owen:
I fixed the documentation as per your suggestions. I am very glad you caught the problem with `allow_multiple_commands` because I renamed that attribute to `allow_multiple_callbacks` and moved it to `ReadTopic` so the documentation needed to be fixed.

PR #100, opened 2020-04-16
DM-24489: Improve the reliability of `tests/test_salobj_to_either.py` by increasing a time limit

r-owen:


r-owen:
I think 99% of the problem is timing. But I don't know what changed and that worries me. I have seen one instance of a different error that appeared to be a dropped message.

I cannot comment on mixing any combination of C/C++/Java. We already know that mixing SAL and salobj in the same process leads to segfaults, but I don't think anyone plans to do that.

PR #101, opened 2020-04-17
DM-23422: improve speed by handling time more efficiently

r-owen:


python/lsst/ts/salobj/base.py
```diff
@@ -230,37 +245,169 @@ def tai_from_utc(utc, format="unix"):
...
+def _update_leap_second_table():
+    """Update the leap second table.
+
+    Notes
+    -----
+    This should be called when this module is loaded.
+    When called, it obtains the current table from AstroPy,
+    then schedules a background (daemon) thread to call itself
+    to update the table ``_LEAP_SECOND_TABLE_UPDATE_MARGIN_DAYS``
+    before the table expires.
```
timj, line 375: So as I understand things, each salobj user is going to ask astropy to update the leap second table, have it written to the standard shared astropy cache as needed, and then read that table in as an internal leap second table?

tests/test_basics.py
```diff
@@ -280,6 +296,9 @@ def check_tai_from_utc(self, utc_ap):
         """
         tai = salobj.tai_from_utc(utc_ap.utc.unix)
 
+        tai_alt = alternate_tai_from_utc_unix(utc_ap.utc.unix)
+        self.assertAlmostEqual(tai, tai_alt, delta=1e-6)
+
```
timj, line 301: Can you add a couple of failing tests? One for 1970 and one for the year 3000? I can't see anywhere where you test that out of range dates will fail.

r-owen: Yes. Excellent suggestion.

python/lsst/ts/salobj/base.py
r-owen: I would express it this way: every time a user imports salobj in Python, salobj it will ask AstroPy for a leap second table. What AstroPy will do depends on how it is configured:
* If AstroPy uses its own table, and its own table is out of date, it will download a new table. Otherwise it will return a copy of the cached table.
* If AstroPy is configured to use the table maintained by NTP or PTP then it will return that table. If that table is out of date we're in trouble, but I think we can choose whether to allow AstroPy to download a new table at that point.

PR #102, opened 2020-04-28
DM-24637: Make CscCommander run commands asynchronously

r-owen:


tests/data/topic_writer.py
```diff
@@ -0,0 +1,103 @@
...
+                else:
+                    self.write_task = asyncio.create_task(self.write_arrays())
+        else:
+            self.write_task.cancel()
+
+    async def write_arrays(self):
+        i = 0
+        while True:
+            self.tel_arrays.data.int0[0] = i
+            self.tel_arrays.put()
```
tribeiro, line 89: Why not use `set_put`? is this faster?

tribeiro: never mind... I see you only want to set the first item...

tests/test_speed.py
```diff
@@ -59,7 +66,7 @@ async def test_readwrite_speed(self):
         This isn't really a unit test, but running it as a test
         avoids bitrot.
         """
-        ntowrite = 500
+        ntowrite = 1000
```
tribeiro, line 69: picky comment could it be `n_to_write` instead? I read it "not write" the first time.. hehe

r-owen: I'll change it to num_to_write

PR #103, opened 2020-05-05
DM-24558: Add SQuaSH metrics for speed

r-owen:


doc/revision_history.rst
```diff
@@ -9,12 +9,24 @@ Revision History
 v5.11.0
 =======
 
-Changes:
+Major changes:
```
tribeiro, line 12: Are they really major changes? This seems like it would fit in a patch release.

python/lsst/ts/salobj/sal_info.py
```diff
@@ -37,12 +37,15 @@
...
-# override by setting env var $LSST_DDS_HISTORYSYNC
+# We want DDS logMessage messages for at least INFO level messages
+# so if the current level is less verbose, set it to INFO.
+# Do not change the level if it is already more verbose,
+# because somebody has intentionally increased verbosity
+# (a common thing to do in unit tests).
+MAX_LOG_LEVEL = logging.INFO
+
+# Default time to wait for historical data (sec);
+# override by setting env var $LSST_DDS_HISTORYSYNC.
```
tribeiro, line 48: It really baffles me that we need a 1 minutes delay to read historical data.

tests/test_speed.py
```diff
@@ -41,11 +47,69 @@ class SpeedTestCase(asynctest.TestCase):
...
+    @classmethod
+    def setUpClass(cls):
+        if verify is None:
+            return
+        metrics = (
+            verify.Metric(
+                name="salobj.CreateClasses",
+                description="The number of topic classes that dds can create "
+                "(from an IDL file), per second. "
+                "This is measured by creating one of of each Test topic. "
```
tribeiro, line 59: double "of" 

> This is measured by creating one of ~of~ each Test topic.

doc/revision_history.rst
r-owen: I believe it is necessary because I added a new function (assert_black_formatted) and added important features to CscCommander.

python/lsst/ts/salobj/sal_info.py
r-owen: That's DDS for you. It takes awhile for a new Domain Participant to "align" itself with the rest of the system. I hope we can reduce this a bit by setting masterPriority (as SAL 4.2 will do) but I doubt it will help enough to reduce this.

Two things to keep in mind:
* This is an upper limit; if late joiner data arrives early then the SAL components will start quickly.
* If we give up too early then we will get stale data masquerading as new data, which is a big problem (especially for commands).

I hope setting masterPriority will lower startup time for scripts, and startup time for components in the full system (once we increase masterPriority for some component that is started early). But even if it does, a long delay is useful for unit tests, where there is no existing master.

tests/test_speed.py
r-owen: Dang. I missed this comment before I merged. I'll fix it now and it'll get into the next version.

PR #104, opened 2020-05-08
Remove ignore flag as no longer necessary

couger01:


conda/meta.yaml
```diff
@@ -37,6 +37,7 @@ test:
     - moto
     - ts-dds
     - ts-idl
+    - salpy-test
```
tribeiro, line 40: where is this package hosted? I though we agreed to move the package to anaconda.

tribeiro, line 40: Also, I would recommend renaming this to `ts-salpy-test`

tribeiro:
We agreed to close this PR without merging. Look at https://jira.lsstcorp.org/browse/DM-24677 for details.

PR #106, opened 2020-05-11
DM-21196: Raise an exception if one tries to read data for a topic before the read loop is running.

r-owen:


tests/test_speed.py
```diff
@@ -137,190 +163,107 @@ async def test_class_creation_speed(self):
...
-                if verify is not None:
-                    self.insert_measurement(
-                        verify.Measurement(
-                            "salobj.ReadTest_logLevel",
-                            log_level_read_speed * u.ct / u.second,
-                        )
-                    )
+            dt = time.monotonic() - t0
+            log_level_read_speed = num_samples / dt
+            nlost = data.level - num_samples
```
mareuter, line 235: Is data.level fixed or incremented during use above? I don't see anything that looks like it getting incremented above and I didn't think python had log levels higher than 50. This looks like I would always see nlost < 0. What am I missing?

tests/test_speed.py
```diff
@@ -137,190 +163,107 @@ async def test_class_creation_speed(self):
...
-                if verify is not None:
-                    self.insert_measurement(
-                        verify.Measurement(
-                            "salobj.ReadTest_arrays",
-                            arrays_read_speed * u.ct / u.second,
-                        )
-                    )
+            dt = time.monotonic() - t0
+            arrays_read_speed = num_samples / dt
+            nlost = data.int0[0] - num_samples
```
mareuter, line 201: This looks like it is comparing a fixed number in an array instead of something that's been counted to num_samples. If data.int0[0] = 1000, nlost is always 0. What am I missing?

doc/revision_history.rst
```diff
@@ -6,6 +6,44 @@
...
+      Instead, print the number of samples lost.
+    * Improve the measurement "salobj.ReadTest_logLevel" by ignoring an extra logLevel event output by `Controller`.
+    * Be more careful about shutting down the topic writer subprocess.
+      This eliminates a warning about an unclosed socket.
+    * Reduced the number of samples read and written, since it doesn't affect the measurements,
+      speeds up the test, and may reduce lost samples.
+    * Removed the combined read/write speed test because it is redundant with the tests added in v5.11.0.
+* Fixed flake8 warnings about f strings with no {}.
+
+Requirements:
```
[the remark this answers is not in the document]
r-owen: Yes they are minimum requirements.

tests/test_speed.py
r-owen: I am using the logLevel message as a representative "small topic" (one of smallest available to me). The data being sent is not a log level, but an incrementing integer.

Writing DDS samples is faster than reading, so with enough samples I would expect the read queues to overfill and messages to be lost. A symptom of that is that some of the incremented values will be lost, and so there will be gaps in the values read. For example, suppose the read values are 0, 1, 2, 3, 7, 8, 9 (in reality I would expect losses to start at much higher values). I measure samples lost as the difference between the final read value and the number of samples read; in this case 9 - 6 = 3 messages lost (the values 4, 5 and 6).

tests/test_speed.py
r-owen: See my comment just above (in the conversation view). This is the same thing being measured as for the logLevel messages, but for a large topic (one with arrays, though I only set the first item in the int0 array).

r-owen:
@tribeiro  can you please review the last few commits, starting with "Modify assert_black..."?

Changes:
* I fixed the issue you had with the black test complaining about version.py (as far as I can tell this only occurs for an installed version, not a git repo, so I had some trouble reproducing it).
* Fixed the .travis.yml file; this suddenly be needed to pass github Travis
* A minor enhancement of CscCommander.

PR #107, opened 2020-05-14
DM-24932: improve output from salobj<->salpy tests

r-owen:


r-owen:
60 seconds is far more than we need for most situations. However, it is proving difficult to judge reasonable values that work on all systems, including heavily loaded CI systems. Rather than try to guess, I felt it better to use one time limit that is definitely long enough.

The time limits are intended to abort a test that is wedged. They are not intended as a sanity check of reasonable times. Originally I had hoped they could be reasonable times, but experience has shown that timings vary wildly, and failing merely because a timeout was a few seconds too short is not useful.

Another option is to get rid of most of the time limits entirely. But the failure mode is tests hanging indefinitely, which is really bad.

PR #108, opened 2020-05-19
Better OS error handling while working with files

pkubanek:
Removes isfile() etc. calls on paths and handles OS problems with try..catch
blocks. This catches corner cases such as file not available for reading.
Discovered during work on DM-24610, but I can make an extra tickets if
preferred.

### DIFF
--- a/python/lsst/ts/salobj/configurable_csc.py
+++ b/python/lsst/ts/salobj/configurable_csc.py
@@ -46,7 +46,7 @@ class ConfigurableCsc(BaseCsc, abc.ABC):
         Name of SAL component.
     index : `int` or `None`
         SAL component index, or 0 or None if the component is not indexed.
-    schema_path : `str`
+    schema_path : `str` or `pathlib.Path`
         Path to a schema file used to validate configuration files
         The recommended path is ``<package_root>/"schema"/f"{name}.yaml"``
         for example:
@@ -131,18 +131,18 @@ class ConfigurableCsc(BaseCsc, abc.ABC):
         initial_simulation_mode=0,
     ):
 
-        if not pathlib.Path(schema_path).is_file():
-            raise ValueError(f"schema_path={schema_path} is not a file")
-        with open(schema_path, "r") as f:
-            schema_data = f.read()
         try:
-            schema = yaml.safe_load(schema_data)
+            with open(schema_path, "r") as f:
+                schema = yaml.safe_load(f.read())
             self.config_validator = DefaultingValidator(schema=schema)
             title = schema["title"]
             name_version = title.rsplit(" ", 1)
             if len(name_version) != 2 or not name_version[1].startswith("v"):
                 raise ValueError(f"Schema title {title!r} must end with ' v<version>'")
             self.schema_version = name_version[1]
+
+        except OSError as e:
+            raise ValueError(f"Could not read {schema_path}: {e}")
         except Exception as e:
             raise ValueError(f"Schema {schema_path} invalid") from e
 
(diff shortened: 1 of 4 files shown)


python/lsst/ts/salobj/configurable_csc.py
```diff
@@ -272,11 +269,12 @@ def _make_config_label_dict(self):
         The result should be put in self.config_label_dict
         """
         labels_path = self.config_dir / "_labels.yaml"
-        if not labels_path.is_file():
-            self.log.warning(f"{labels_path} not found")
+        try:
+            with open(labels_path, "r") as f:
+                labels_raw = f.read()
+        except OSError as e:
+            self.log.warning(f"{labels_path} not found: {e}")
```
r-owen, line 276: "not found" seems like the wrong message, since there might be other errors. Please consider something like:
```
f"Could not read labels {labels_path}: {e}"
```

python/lsst/ts/salobj/configurable_csc.py
```diff
@@ -372,13 +370,11 @@ async def begin_start(self, data):
...
-                        f"Cannot find config file {config_file_name} "
-                        f"in {self.config_dir}"
-                    )
-                with open(config_file_path, "r") as f:
-                    config_yaml = f.read()
+                try:
+                    with open(config_file_path, "r") as f:
+                        config_yaml = f.read()
+                except OSError as e:
+                    raise base.ExpectedError(f"Cannot read {config_file_path}: {e}")
```
r-owen, line 377: Please provide context by saying what the file is that it can't read, e.g.: `f"Cannot read config file {config_file_path}: {e}"`

python/lsst/ts/salobj/configurable_csc.py
```diff
@@ -131,18 +131,18 @@ def __init__(
...
+                schema = yaml.safe_load(f.read())
             self.config_validator = DefaultingValidator(schema=schema)
             title = schema["title"]
             name_version = title.rsplit(" ", 1)
             if len(name_version) != 2 or not name_version[1].startswith("v"):
                 raise ValueError(f"Schema title {title!r} must end with ' v<version>'")
             self.schema_version = name_version[1]
+
+        except OSError as e:
+            raise ValueError(f"While reading schema_path={schema_path}: {e}")
```
r-owen, line 145: I strongly suggest a message such as `f"Could not read schema {schema_path}: {e}"`
as I think it would be clearer.

r-owen: Also, if you are willing, please fix an existing error: schema_path is described as a "str" in the doc string, but it should be described as:
```
schema_path : `str` or `pathlib.Path`
```
It is possible to be even more general, but the above is what I have been using.

python/lsst/ts/salobj/configurable_csc.py
```diff
@@ -158,10 +158,7 @@ def __init__(
             config_dir = self.get_default_config_dir()
         else:
             config_dir = pathlib.Path(config_dir)
-            if not config_dir.is_dir():
-                raise ValueError(
-                    f"config_dir={config_dir} does not exists or is not a directory"
-                )
+
```
r-owen, line 161: Please restore this code. This catches a common error early, before anyone tries to configure the CSC.

r-owen: Please restore this code.

python/lsst/ts/salobj/configurable_csc.py
```diff
@@ -428,7 +431,8 @@ def get_config_pkg():
         """
         raise NotImplementedError()
 
-    def get_default_config_dir(self):
+    # was public - def get_default_config_dir(self):
+    def __get_default_config_dir(self):
```
r-owen, line 435: Please just use one leading underscore. I prefer not to invoke the Python name munging mechanism unless it is truly necessary. I would rather rely on users to be sensible, and allow overriding a method if a user has a good reason to do so.

python/lsst/ts/salobj/configurable_csc.py
```diff
@@ -428,7 +431,8 @@ def get_config_pkg():
         """
         raise NotImplementedError()
 
-    def get_default_config_dir(self):
+    # was public - def get_default_config_dir(self):
```
r-owen, line 434: Please delete this comment.

PR #114, opened 2020-05-27
Tickets/dm 25073: add create and profile arguments to AsyncS3Bucket's constructor

r-owen:


r-owen:
The CI breakage is due to adding a new generic command to the XML (on develop, which surprises me; I thought CI used master). I may as well fix it on this ticket branch.

PR #115, opened 2020-06-01
Tickets/dm 25179: add angle wrapping functions

r-owen:


doc/version_history.rst
```diff
@@ -6,6 +6,22 @@
...
 ###############
 
+v5.15.0
+=======
+
+Changes:
+
+* Add `angle_wrap_center` and `angle_wrap_nonnegative` functions.
+* Broke the test of black formatting out into its own test file ``test_black.py``,
+  to make it easies to copy into other packages.
```
wvreeven, line 16: There is a typo in line 16. "easies" should be "easier".

python/lsst/ts/salobj/base.py
```diff
@@ -130,6 +138,59 @@ class ExpectedError(Exception):
...
+
+    Parameters
+    ----------
+    angle : `astropy.coordinates.Angle` or `float`
+        Angle; if a float then in degrees
+
+    Returns
+    -------
+    wrapped : `astropy.coordinates.Angle`
+        angle wrapped into the range -180 <= wrapped < 180 deg.
```
wvreeven, line 173: In line 163 you define the range as [-180, 180] while here you define it as 180 <= wrapped < 180.

PR #117, opened 2020-06-10
DM-25356: Update a unit test for compatibility with ts_sal 4.2

r-owen:


r-owen:
I am not sure why Jenkins failed. It looks like the same problem I was having: no communication between salobj and SAL, until I updated my ts_sal develop to commit d808fecf6b61, which is current as I write this. I saw the problem with a slightly earlier commit.

PR #118, opened 2020-06-15
Cleanup Jenkins workspace after conda/pypi build.

tribeiro:


pkubanek:
Consider instead to let Jenkins clear the workspace. That can be achieved by adding:

```
properties(
    [
    buildDiscarder
        (logRotator (
            artifactDaysToKeepStr: '',
            artifactNumToKeepStr: '',
            daysToKeepStr: '14',
            numToKeepStr: ''
        ) ),
    disableConcurrentBuilds()
    ]
)
```

on top (before node/pipeline). That will clear after 14 days (can be changed). Check with me or @rbovill for details. The advantage of this approach is that your build environment will be preserved for possible debugging.

rbovill:
I added in the DisableConcurrentBuilds earlier, because this job tended to be quite noisy, and would occasionally cause traffic jams.

PR #120, opened 2020-06-16
Changed to scripted pipeline

pkubanek:
Use Jenkins/Groovy docker plugin. Makes script a bit more readable. Build in workspace mounted inside container.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,28 +1,31 @@
-pipeline {
-    agent any
-    environment {
-        network_name = "n_${BUILD_ID}_${JENKINS_NODE_COOKIE}"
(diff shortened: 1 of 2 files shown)


rbovill:
Why are we moving back to scripted pipeline syntax?  Declarative syntax is the standard format now.

pkubanek:
> Why are we moving back to scripted pipeline syntax? Declarative syntax is the standard format now.

You are right. I will rework it back to declarative - my fault, I need scripted for something else.

rbovill:
Closing this PR as OBE.  Will try again in the future with Declarative Syntax, if desired.

PR #126, opened 2020-07-07
DM-25825: add the get_opensplice_version function

r-owen:


python/lsst/ts/salobj/base.py
```diff
@@ -78,6 +80,9 @@
 # Regex for a SAL componet name encoded as <name>[:<index>]
```
wvreeven, line 80: An existing comment but please correct the spelling of component.

PR #127, opened 2020-07-13
DM-25912: get rid of automatic CMD_INPROGRESS ack

r-owen:


doc/salobj_cscs.rst
```diff
@@ -101,7 +101,7 @@ Most CSCs can be configured.
       * ``setValue``. This is strongly discouraged, for reasons given below.
```
tribeiro, line 101: I think this `CSC Details` would benefit from some reformatting. The information there is very dense and does't present in a good way to the readers. 

I think you could make each of those major bullet points into subsections with an explanatory title and some support text. I would also like to see an example of how to issue the acknowledgement. I am sure I could figure out by doing some digging around based on the information here but I think it would be better if **all** the information was present here.

r-owen: We might want to make that a separate ticket.

tribeiro: agreed....

python/lsst/ts/salobj/topics/controller_command.py
```diff
@@ -209,19 +218,18 @@ async def _run_callback(self, data):
         try:
             result = self._callback(data)
             if inspect.isawaitable(result):
-                self.ackInProgress(data)
                 ack = await result
             else:
                 ack = result
             if ack is None:
-                ack = self.salinfo.makeAckCmd(
+                ack = self.salinfo.make_ackcmd(
```
tribeiro, line 225: Uhm... did I understood this correctly that `ack` is the return value of the `do_<command>` method? what happens if the someone inadvertently returns something? the command will o be acknowledged?

wouldn't it be cleaner/safer to just do 

```
            if inspect.isawaitable(result):
                await result

             ack = self.salinfo.make_ackcmd(
                    private_seqNum=data.private_seqNum,
                    ack=sal_enums.SalRetCode.CMD_COMPLETE,
                    result="Done",
                )
```

r-owen: There are 4 cases:
* do_<command> returns None. The command is reported as successful
* do_<command> raises. The command is reported as failed
* do_<command> returns an `ackcmd` data sample. That is used as the final ack for that command. This is *very unusual* and is only used when the do_<command> method wants to exit, but trigger some other background process that finishes processing the command.
* do_<command> returns anything else. This will cause an exception and the command will be reported as failed.

None of this has changed at all in this commit.

r-owen: I will expand the docs for ControllerCommand a bit.

PR #129, opened 2020-07-31
DM-26191: Improve docs for CscCommander.amain

r-owen:
and make it obvious that index is a required argument.

### DIFF
--- a/doc/version_history.rst
+++ b/doc/version_history.rst
@@ -32,6 +32,7 @@ Changes:
 
 * Implemented authorization support.
   This version will communicate with ts_sal 4.2 and ts_salobj 5, but authorization support will be limited until the whole system uses ts_sal 5 and ts_salobj 6.
+* `CscCommander` now rounds float arrays when displaying events and telemetry (it already rounded float scalars).
 * Added support for running without a durability service:
   set environment variable ``LSST_DDS_HISTORYSYNC`` to a negative value to prevent waiting for historical data.
 * Added the `get_opensplice_version` function.
--- a/python/lsst/ts/salobj/base_csc.py
+++ b/python/lsst/ts/salobj/base_csc.py
@@ -156,8 +156,6 @@ class BaseCsc(Controller):
             If the CSC is not indexed: specify `None` or 0.
         **kwargs : `dict`, optional
             Additional keyword arguments for your CSC's constructor.
-            If any arguments match those from the command line
-            the command line values will be used.
 
         Returns
         -------
@@ -198,8 +196,6 @@ class BaseCsc(Controller):
             If the CSC is not indexed: specify `None` or 0.
         **kwargs : `dict`, optional
             Additional keyword arguments for your CSC's constructor.
-            If any arguments match those from the command line
-            the command line values will be used.
         """
         csc = cls.make_from_cmd_line(index=index, **kwargs)
         await csc.done_task
--- a/python/lsst/ts/salobj/csc_commander.py
+++ b/python/lsst/ts/salobj/csc_commander.py
@@ -78,9 +78,16 @@ def round_any(value, digits=4):
     """Round any value to the specified number of digits.
 
     This is a no-op for int and str values.
+
+    Notes
+    -----
+    This is designed for use with DDS samples. As such, it expects
+    array values to be of type `list`.
     """
     if isinstance(value, float):
         return round(value, digits)
+    elif value and isinstance(value, list) and isinstance(value[0], float):
+        return [round(item, digits) for item in value]
     return value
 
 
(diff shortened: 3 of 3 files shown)


pkubanek:
I would like to see minor improvements:

* adding an example how to use csc_commander - although it will not work (due to enable being required argument, which is result of mixing variable initialization and assignment in constructor; while Python allows for that and compared to C++ code, I believe in this case it significantly increases complexity of design and we should split constructor into a construct and method which will run things)
* simplifies float rounding

see the diff:

```diff
diff --git a/python/lsst/ts/salobj/csc_commander.py b/python/lsst/ts/salobj/csc_commander.py
index a54efc3..9cb831e 100644
--- a/python/lsst/ts/salobj/csc_commander.py
+++ b/python/lsst/ts/salobj/csc_commander.py
@@ -74,16 +74,6 @@ async def stream_as_generator(stream, encoding="utf-8"):
         yield line
 
 
-def round_any(value, digits=4):
-    """Round any value to the specified number of digits.
-
-    This is a no-op for int and str values.
-    """
-    if isinstance(value, float):
-        return round(value, digits)
-    return value
-
-
 class CscCommander:
     """Command a CSC from the command line.
 
@@ -143,10 +133,16 @@ class CscCommander:
     * evt_<event_name>_callback overrides handling data for the specified
       event (usually this just mean printing the data).
       It receives one argument: the DDS sample.
+    * event_callback is default callback for events, receiving the DDS sample
+      and event name. The class implementation prints events data.
     * tel_<event_name>_callback overrides handling data for the specified
       telemetry topic (usually this just involves printing the data).
       It receives one argument: the DDS sample.
       This can be useful if the default is too chatty.
+    * telemetry_callback is default callback for telemetry data. Receives the
+      DDS sample and telemetry name. The class implementation compares received
+      data with previous data, and if a significant difference is found (float
+      values are compared rounded to 5 digits), prints out the received data.
     * Hide unwanted commands by deleting them from ``self.command_dict``
       in your constructor. The following is appropriate for most CSCs::
 
@@ -158,6 +154,32 @@ class CscCommander:
     how to send multiple commands (the ``suprocess.communicate``
     method only allows sending one item of data).
     Instead I suggest manually running it to control the Test CSC.
+
+    Examples
+    --------
+
+    >>> import asyncio
+    >>> from lsst.ts.salobj import CsCCommander
+    >>> asyncio.run(CsCCommander(name='Test').amain())
+
+    If you would like to customize handling of subsystems commands, telemetry
+    etc, create CsCCommander subclass:
+
+    import asyncio
+
+    class TestCommands(salobj.CsCCommander):
+        def __init__(self):
+            super().__init__(name="Test")
+
+        def event_callback(self, data, name):
+            # Disable printout of all unhandled events
+            pass
+
+        def evt_testInfo_callback(self, data):
+            print(f"TestInfo {str(data)}")
+
+
+    asyncio.run(TestCommands.amain())
     """
 
     def __init__(
@@ -207,7 +229,7 @@ class CscCommander:
         """
         command_help = "\n".join(self.get_commands_help())
         print(
-            f"""Send commands to the {self.remote.salinfo.name} CSC and print events and telemetry.
+            """Send commands to the {self.remote.salinfo.name} CSC and print events and telemetry.
 
 CSC Commands:
 {command_help}
@@ -266,11 +288,11 @@ help  # print this help
             if self.field_is_public(key)
         )
 
-    def get_rounded_public_fields(self, data):
+    def _get_rounded_public_fields(self, data):
         """Get the public fields for a sample, with float values rounded.
         """
         return {
-            key: round_any(value)
+            key: round(value, 5) if isinstance(value, float) else value
             for key, value in data.get_vars().items()
             if self.field_is_public(key)
         }
@@ -295,9 +317,15 @@ help  # print this help
 
         You may provide tel_<telemetry_name> methods to override printing
         of specific telemetry topics.
+
+        Parameters
+        ----------
+        data : CsC data
+
+        name :
         """
         prev_value_name = f"previous_tel_{name}"
-        public_fields = self.get_rounded_public_fields(data)
+        public_fields = self._get_rounded_public_fields(data)
         if public_fields != getattr(self, prev_value_name):
             setattr(self, prev_value_name, public_fields)
             formatted_data = ", ".join(
@@ -452,7 +480,7 @@ help  # print this help
                 self.remote.evt_summaryState.callback = summary_state_callback
 
     @classmethod
-    async def amain(cls, *, index, **kwargs):
+    async def amain(cls, index=0, enable=False, **kwargs):
         """Construct the commander and run it.
 
         Parse the command line to construct the commander,
@@ -460,11 +488,12 @@ help  # print this help
 
         Parameters
         ----------
-        index : `int`, `True`, `False` or `None`
+        index : `int` or `None`, optional
             If the CSC is indexed: specify `True` make index a required
             command line argument, or specify a non-zero `int` to use
             that index.
-            If the CSC is not indexed: specify `None` or 0.
+            If the CSC is not indexed: specify `None` or 0. Default is None.
+        enable : `True`, `False` or `None, optional
         **kwargs : `dict`, optional
             Additional keyword arguments for your CSC's constructor.
         """
```

r-owen:
`enable` should *not* be an argument to `amain` because `amain` calls `make_from_cmd_line` which is where the `enable` argument comes from.

I prefer the rounding code as is because I can expand the function to add array support more easily than trying to do that inline. In fact I plan to push that change to this ticket branch shortly.

The rounding method is public because I expect subclasses may wish to override it.

I will add an examples section.

PR #132, opened 2020-08-14
Update the version history for v5.17.1

r-owen:


couger01:
Would you be willing to remove the `.travis.yml` file? I just think its redundant given that our Jenkins CI runs flake8 checks as part of the unit tests. Also would reduce maintenance overhead.

PR #141, opened 2020-09-01
DM-26556: Implement the new DDS partition naming scheme

r-owen:


python/lsst/ts/salobj/base_csc.py
```diff
@@ -190,7 +190,12 @@ def make_from_cmd_line(cls, index, **kwargs):
         """
         parser = argparse.ArgumentParser(f"Run {cls.__name__}")
         if index is True:
-            parser.add_argument("index", type=int, help="Script SAL Component index.")
+            parser.add_argument("index", type=int, help="SAL index.")
+        parser.add_argument(
+            "--loglevel",
+            type=int,
+            help="log level: error=40, warning=30, info=20, debug=10",
```
tribeiro, line 197: Consider using `choices = [40, 30, 20, 10]` or maybe even naming the levels and parsing them afterwards. Unless the intention is to support non standard levels.

python/lsst/ts/salobj/configurable_csc.py
```diff
@@ -480,7 +480,7 @@ def _get_default_config_dir(self):
     def add_arguments(cls, parser):
         parser.add_argument(
             "--configdir",
-            help="Directory containing configuration files for the start command.",
+            help="directory containing configuration files for the start command.",
```
tribeiro, line 483: out of curiosity, why "directory" instead of "Directory"?

python/lsst/ts/salobj/sal_info.py
```diff
@@ -227,19 +249,17 @@ def __init__(self, domain, name, index=0):
...
-
-        publisher_qos = domain.qos_provider.get_publisher_qos()
-        publisher_qos.set_policies([partition_qos_policy])
-        self.publisher = domain.participant.create_publisher(publisher_qos)
-
-        subscriber_qos = domain.qos_provider.get_subscriber_qos()
-        subscriber_qos.set_policies([partition_qos_policy])
-        self.subscriber = domain.participant.create_subscriber(subscriber_qos)
+        self.partition_prefix = os.environ.get("LSST_DDS_PARTITION_PREFIX")
+        if self.partition_prefix is None:
```
tribeiro, line 253: I think it would be good to add a warning if `LSST_DDS_DOMAIN` is set even if `LSST_DDS_PARTITION_PREFIX` is also set. Just to make sure people is aware of what is happening.

python/lsst/ts/salobj/sal_info.py
```diff
@@ -251,6 +271,12 @@ def __init__(self, domain, name, index=0):
         self.authorized_users = set()
         self.non_authorized_cscs = set()
 
+        # Publishers and subscribers; create at need to avoid
```
tribeiro, line 274: to avoid what?

tribeiro (approved):
Changes look good. I appreciate the deprecation warnings. I made a couple comments inline, feel free to adopt what you think is most sensible.

python/lsst/ts/salobj/base_csc.py
r-owen: The intent is to support all levels.

python/lsst/ts/salobj/configurable_csc.py
r-owen: to match the help text for built-in arguments such as `--help` and `--version`. I only recently realized they start with lowercase.

python/lsst/ts/salobj/sal_info.py
r-owen: How about if I warn only if they are different?

python/lsst/ts/salobj/sal_info.py
r-owen: Oops. I fixed the comment and also expanded some related material.

PR #143, opened 2020-09-10
DM-26664: use the new QoS file with separate profiles for each topic category.

r-owen:


python/lsst/ts/salobj/sal_info.py
```diff
@@ -752,6 +731,10 @@ async def _read_loop(self):
                             condition, reader._data_queue.maxlen
                         )
                         not reader.dds_queue_length_checker.check_nitems(len(data_list))
+                        if self.debug:
```
tribeiro, line 734: Is this really necessary? Consider adding why not add a `log.debug` statement instead? if you want to skip the loop when level is larger than debug you could do:

```
if self.log.level < logging.INFO:
    ...
```

python/lsst/ts/salobj/sal_info.py
```diff
@@ -236,6 +236,7 @@ def __init__(self, domain, name, index=0):
             raise TypeError(f"domain {domain!r} must be an lsst.ts.salobj.Domain")
         self.isopen = True
         self.domain = domain
+        self.debug = False
```
tribeiro, line 239: Consider not using this attribute and determining if running in debug mode from the log level. I think it will make the code cleaner.

python/lsst/ts/salobj/domain.py
```diff
@@ -43,6 +43,55 @@
 MAX_RANDOM_HOST = (1 << 31) - 1
```
tribeiro, line 43: The docstring for `DDS_READ_QUEUE_LEN` above says:

```
his must be equal to or greater than the queue length in the
OpenSplice configuration file (pointed to by $OSPL_URI).
```

But it is not in `OSPL_URI` right? Isn't it in the QoS file?

Also, is it still true that you can't read it from the QoS objects you create bellow? If you can read it from these objects created from the QoS file, then you can probably ditch this value here.

python/lsst/ts/salobj/domain.py
```diff
@@ -43,6 +43,55 @@
...
+            CommandProfile=True,
+            EventProfile=False,
+            TelemetryProfile=True,
+        )[profile_name]
+
+        self.qos_provider = dds.QosProvider(qos_path.as_uri(), profile_name)
+        self.topic_qos = self.qos_provider.get_topic_qos()
+
+        read_queue_policy = dds.HistoryQosPolicy(
+            depth=DDS_READ_QUEUE_LEN, kind=dds.DDSHistoryKind.KEEP_LAST
```
tribeiro, line 87: I really hope you can get `DDS_READ_QUEUE_LEN` from the QoS file.

tribeiro: Ok, I think I am missing something here. When we change the length of the read queue are we just changing the length of the python queue? I thought we should be changing the size of the DDS read queue too. If that is not the case, I don't think it makes much sense to have this feature.

tribeiro (commented):
I made a few comments in line. I hope you will consider some of them. In addition  I just want to make sure I understand the read queue length feature before continuing.

python/lsst/ts/salobj/sal_info.py
r-owen: That was debugging code accidentally left in. I will remove it.

python/lsst/ts/salobj/sal_info.py
r-owen: Debugging code accidentally left in. I will remove it. (This was part of my trying to figure out why some commands were being lost -- which was the QoS issue you diagnosed).

r-owen: Fixed (and the next item). Thank you for catching this.

python/lsst/ts/salobj/domain.py
r-owen: Yes it is the QoS file. I will fix the doc string.

It is still true that I cannot read any QoS values from the dds library. It is maddening. But I don't understand why I can ditch this value here.

r-owen: Oh, I think I see your point. You mean that I should not have to set read_queue_policy:
```
        read_queue_policy = dds.HistoryQosPolicy(
            depth=DDS_READ_QUEUE_LEN, kind=dds.DDSHistoryKind.KEEP_LAST
        )
        self.reader_qos = self.qos_provider.get_reader_qos()
        self.reader_qos.set_policies([read_queue_policy])
```
I will try removing that code.

python/lsst/ts/salobj/domain.py
r-owen: The point is to choose a value that can safely be used in the `dds.Reader.take_cond`.

r-owen: OK, here's the scoop. I cannot get the read queue length from dds, alas, but it was only really only wanted for to print warnings about the DDS read queues filling up. So I changed this to an attribute of QosSet (which allows it to have a different value for each type of message). It should still be updated to match the values in the QoS file, but if it is not, it only affects the accuracy of the warnings.

The other code that used this constant was changed as follows:
* SalInfo has a new constant `MAX_HISTORY_READ` that sets the maximum number of historical samples to read. I documented it carefully. Being smaller than the maximum length of any DDS write queue is very dangerous because it leads to historical data being mixed with new data. Being too large is no problem. So I set it to 10,000.
* The read topics now have a new constant `DEFAULT_QUEUE_LEN` which sets the default value of the `queue_len` constructor argument, which sets the length of the Python queue (not the DDS read queue). The value is 100, so the default is unchanged.

Domain no longer sets the reader QoS policy, so QoS is now *entirely* determined by the new QoS.xml file. I think this is an excellent change. I was afraid it might break something, but it appears to be fine.

python/lsst/ts/salobj/domain.py
r-owen: I was able to remove it and all tests pass. Thank you for suggesting this change. Now *all* QoS is defined by the new QoS.xml file in ts_idl.

python/lsst/ts/salobj/domain.py
```diff
@@ -31,16 +31,65 @@
...
+    def __init__(self, qos_path, profile_name):
+        self.profile_name = profile_name
+        self.volatile = dict(
+            AckcmdProfile=True,
+            CommandProfile=True,
+            EventProfile=False,
+            TelemetryProfile=True,
+        )[profile_name]
+
+        self.dds_read_queue_len = 100
```
tribeiro, line 87: So, this may be a picky comment but can you consider changing the name of the attribute to something that make it clear it is not the actual length of the queue? maybe

```
self.dds_read_queue_warning_level
```

or something on those lines?

Also, I wonder if this should be an editable attribute. Maybe make this a `@property`, which would allow you to document it even better? 

```
@property
def dds_read_queue_warning_level(self):
    """ ....
    """
    return 100
```

tribeiro: So, as I was saying in private, I think this information is actually accessible through python. 

If you do;

```
self.reader_qos.history.depth
```

you will get the size of the queue.

python/lsst/ts/salobj/topics/read_topic.py
```diff
@@ -240,7 +246,7 @@ def __init__(
         name,
         sal_prefix,
         max_history,
-        queue_len=DDS_READ_QUEUE_LEN,
+        queue_len=DEFAULT_QUEUE_LEN,
```
tribeiro, line 249: I wonder if we should really have a default value for this parameter. I think it is quite a critical and important parameter, and I think you are always setting it when you start no?

python/lsst/ts/salobj/domain.py
```diff
@@ -215,6 +222,40 @@ def add_salinfo(self, salinfo):
...
+        """Make a dds publisher.
+
+        Parameters
+        ----------
+        partition_names : `list` [`str`]
+            List of DDS partition names.
+        """
+        partition_qos_policy = dds.PartitionQosPolicy(partition_names)
+
+        # Any of the qos sets will dom because the publisher and subscriber
```
tribeiro, line 235: Not sure I understood this comment. Is it a typo in "will dom"?

python/lsst/ts/salobj/topics/read_topic.py
```diff
@@ -188,6 +190,10 @@ class ReadTopic(BaseTopic):
         If queue_len <= 0.
     ValueError
         If max_history > queue_len.
+        Note: the DDS write queue length sets an upper limit on how many
+        historical samples are available. However, dds offer no way
```
tribeiro, line 194: I think comments need fixing, we can read the size of the queues thought python, no?

python/lsst/ts/salobj/domain.py
r-owen: It should be "do", not "dom". I'll fix it.

python/lsst/ts/salobj/topics/read_topic.py
r-owen: I think most code uses the default.

python/lsst/ts/salobj/topics/read_topic.py
r-owen: Fixed

python/lsst/ts/salobj/domain.py
r-owen: Fixed

python/lsst/ts/salobj/domain.py
r-owen: Great suggestion. I have updated the code accordingly. Note that reader_qos.durability_service.history_depth is always reported as 0 rather than the correct value so, much as I'd like to use that value to warn for excessively large values of max_history, I can't. I filed a technical debt ticket for that.

tribeiro (approved):
Alright! I think we are good to go now!

PR #144, opened 2020-09-16
Attempt to understand and add clarity to ReadTopic documentation

patrickingraham:
I've recently got myself wrapped around a poll trying to understand the different usages of ReadTopic. I think this could be partially cleared up with increased documentation. However, I don't fully understand it at a level where I can simply write it and submit a PR.
Therefore, please note that the current state of this PR is more for discussion. I fully expect to get clarity on a few issues then we can massage the text some more and hopefully help clear this up for others.

A note for context... i somewhat got into the tailspin because I was expecting the following to work during a test i was writing the following and expecting it to return the next message each time it was run in a loop, when it only returns the first value published
```
Controller.evt_YYY.set_put(example_attribute='b')
remote.evt_YYY.flush()
remote.evt_YYY.aget(timeout=5)
```
I then switched to using `next` but then later on wanted to grab that same message later, but using `next` popped it off so it wasn't there when running `next` with flush=False. I worked around it (and made the code better in doing so), but I still find the behaviour confusing...

### DIFF
--- a/python/lsst/ts/salobj/topics/read_topic.py
+++ b/python/lsst/ts/salobj/topics/read_topic.py
@@ -434,7 +434,11 @@ class ReadTopic(BaseTopic):
         self._data_queue.clear()
 
     async def aget(self, timeout=None):
-        """Get the current message, if any, else wait for the next message.
+        """Get the most recent message, if any, else wait for the next message
+        to appear.
+        This routine is *not* a pop operation (see `next`) and leaves the
+        queue contents unaltered.
+        It is also unaffected by the `flush` method.
 
         Parameters
         ----------
@@ -466,6 +470,7 @@ class ReadTopic(BaseTopic):
 
     def flush(self):
         """Flush the queue of unread data.
+        This flush will not affect the output of `aget`.
 
         Raises
         ------
@@ -478,6 +483,13 @@ class ReadTopic(BaseTopic):
 
     def get(self, flush=True):
         """Get the most recently seen message, or `None` if no data ever seen.
+
+        This routine is *not* a pop operation (see `next`) but will flush
+        the queue (by default), which affects the `next` method.
+
+        This routine is also unaffected by the `flush` method.
+
+        Use the `aget` function if wanting to wait for the next message to arrive.
 
         Parameters
         ----------
@@ -529,7 +541,13 @@ class ReadTopic(BaseTopic):
         return None
 
     async def next(self, *, flush, timeout=None):
-        """Wait for a message, possibly returning the oldest queued message.
+        """Wait for a message, or immediately return the oldest queued message.
+        This is a pop operation and the message will be permanently removed
+        from the queue.
+
+        To ensure the next message is received, the topic should be flushed
+        prior to being populated using the `flush` method, then this method
+        should be called/awaited with the flush attribute set to False
 
         Parameters
         ----------


python/lsst/ts/salobj/topics/read_topic.py
```diff
@@ -529,7 +541,13 @@ def get_oldest(self):
...
+        """Wait for a message, or immediately return the oldest queued message.
+        This is a pop operation and the message will be permanently removed
+        from the queue.
+
+        To ensure the next message is received, the topic should be flushed
+        prior to being populated using the `flush` method, then this method
+        should be called/awaited with the flush attribute set to False
 
         Parameters
         ----------
```
patrickingraham, line 553: I don't understand the "if any" case below. You're saying that regardless of how the `flush` parameter is set the routine will wait for a new message before returning? Even if it returns the oldest message (which would have already been in the queue?

python/lsst/ts/salobj/topics/read_topic.py
```diff
@@ -466,6 +470,7 @@ async def aget(self, timeout=None):
 
     def flush(self):
         """Flush the queue of unread data.
+        This flush will not affect the output of `aget`.
```
patrickingraham, line 473: Does it affect the output of `get` ? Presumably not?

python/lsst/ts/salobj/topics/read_topic.py
```diff
@@ -479,6 +484,13 @@ def flush(self):
     def get(self, flush=True):
         """Get the most recently seen message, or `None` if no data ever seen.
 
+        This routine is *not* a pop operation (see `next`) but will flush
```
patrickingraham, line 487: I suppose I'm rather confused regarding why the `get` method and the `aget` method differ regarding their behaviour/presence of the `flush` attribute.

It feels like the flush really shouldn't be here, and it appears to be here because of backwards compatibility. Is it getting deprecated soon?

patrickingraham:
A note for context... i somewhat got into the tailspin because I was expecting the following to work during a test i was writing the following and expecting it to return the next message each time it was run in a loop, when it only returns the first value published
```
Controller.evt_YYY.set_put(example_attribute='b')
remote.evt_YYY.flush()
remote.evt_YYY.aget(timeout=5)
```
I then switched to using `next` but then later on wanted to grab that same message later, but using `next` popped it off so it wasn't there when running `next` with flush=False. I worked around it (and made the code better in doing so), but I still find the behaviour confusing...

python/lsst/ts/salobj/topics/read_topic.py
r-owen: Correct. It affects neither of them. In https://github.com/lsst-ts/ts_salobj/pull/145 I changed the brief description to:

"Flush the queue used by `get_oldest` and `next`."

and left it at that. If you want more text, such as an explicit statement that data returned by `get` and `aget` is not affected, I can add that (but perhaps look at the new doc strings for those methods, first).

python/lsst/ts/salobj/topics/read_topic.py
r-owen: Good question.

The `flush` argument for `get` is a holdover from the very early days of salobj. If I was starting over I would either omit, require it, or change the default to False. Given all the existing code the safest thing to do is require it, and https://github.com/lsst-ts/ts_salobj/pull/145 is the first step: issue a deprecation warning if the argument is omitted.

Rather than repeat that mistake for `aget` (a more recent addition) I omitted the `flush` argument.

python/lsst/ts/salobj/topics/read_topic.py
r-owen: The new brief string for `next` is  "Pop and return the oldest message from the queue, waiting for data if the queue is empty." on https://github.com/lsst-ts/ts_salobj/pull/145/files.

r-owen (commented):
I made a few comments in line. I know you have already done this once, but please comment on https://github.com/lsst-ts/ts_salobj/pull/145/files for any further clarifications you would like to see.

PR #145, opened 2020-09-17
DM-26772: deprecate omitting the flush argument of get

r-owen:


python/lsst/ts/salobj/topics/read_topic.py
```diff
@@ -215,35 +215,36 @@ class ReadTopic(BaseTopic):
...
 
     **Modifying Messages**
 
-    All functions that return messages return from an internal cache.
-    This cached value is replaced by all read operations. So as long as
-    no other code modifies a message, the returned message will not change.
-
-    To safely modify a returned message, make your own copy
-    with ``copy.copy(data)``.
+    All functions that return messages return them from an internal cache.
```
patrickingraham, line 244: How is the cache different from the Python queue itself?

python/lsst/ts/salobj/topics/read_topic.py
```diff
@@ -476,15 +482,22 @@ def flush(self):
...
         self._data_queue.clear()
 
-    def get(self, flush=True):
-        """Get the most recently seen message, or `None` if no data ever seen.
+    def get(self, flush=None):
+        """Get the most recent message, or `None` if no data available
+        (`has_data` False).
+
+        This method does not read from the queue, though it optionally
+        flushes the queue.
```
patrickingraham, line 490: Maybe add that this functionality will be deprecated in the future?

python/lsst/ts/salobj/topics/read_topic.py
```diff
@@ -434,7 +435,10 @@ async def close(self):
         self._data_queue.clear()
 
     async def aget(self, timeout=None):
-        """Get the current message, if any, else wait for the next message.
+        """Get the most recent message, or wait for data if none available
+        (`has_data` False).
+
+        This method neither reads from, nor flushes, the queue.
```
patrickingraham, line 441: Then it comes from the cache?

python/lsst/ts/salobj/topics/read_topic.py
```diff
@@ -531,6 +550,8 @@ def get_oldest(self):
     async def next(self, *, flush, timeout=None):
         """Wait for a message, possibly returning the oldest queued message.
 
+        This reads and pops an item from the queue.
```
patrickingraham, line 553: but the item is still available from `get` (or `aget` with flush=False, yes?)
Is it correct to say,
"This reads and pops an item from the queue, but leaves the cache (used by `get` and `aget` unaltered)"

python/lsst/ts/salobj/topics/read_topic.py
```diff
@@ -215,35 +215,36 @@ class ReadTopic(BaseTopic):
     -----
     **Queues**
 
-    There are actually two queues: a Python queue whose length
-    is set by ``queue_len`` and a dds queue whose length is set by
-    the DDS QoS file. Data can be lost in two ways:
+    There are two queues: a Python queue whose length is set by ``queue_len``
```
patrickingraham, line 218: I think a sentence or two on why both exist might be helpful?
"The Python queue exists to enable ..."

patrickingraham (changes requested):
Certainly improved, but I wonder if explanation of the `cache` and how it is used by `aget` and `get` (if I understood correctly)?
See comments in files for more details.

python/lsst/ts/salobj/topics/read_topic.py
r-owen: What is deprecated is calling `get` without specifying the `flush` argument.

I could be convinced to deprecate allowing `get` to flush the queue at all. But the code and documentation would have to be written differently to do that. Being able to flush is occasionally convenient because it's a "best effort" and is allowed (but ignored) if there is a callback function, whereas the `flush` method raises an exception if there is a callback. Still...users should know whether there is a callback or not so it's not that big a deal.

python/lsst/ts/salobj/topics/read_topic.py
r-owen: There are at least two sources of cached data: the Python queue and also the `_current_data` attribute.

python/lsst/ts/salobj/topics/read_topic.py
r-owen: It comes from the `_current_data` attribute, which is one source of cached data, the queue being the other.

python/lsst/ts/salobj/topics/read_topic.py
r-owen: Good point. Yes it does not affect the value returned by aget or get. I will add that information.

r-owen:
@patrickingraham I took another pass at the ReadTopic documentation, trying to focus on the effect each data reading method has on the other methods. I hope that will help clarify what's going on.

python/lsst/ts/salobj/topics/read_topic.py
r-owen: I pushed another change with that info.

python/lsst/ts/salobj/topics/read_topic.py
r-owen: Good question. I have reworded the text of all data reading methods to make it explicit what affect they have on other data reading methods. Please have another look.

python/lsst/ts/salobj/topics/read_topic.py
r-owen: Good suggestion. I added this as another commit. I would love to just use the dds queue but between the severely limited API and lack of asyncio support it did not seem practical.

patrickingraham (approved):
Much better, thank you.
I think some of the behaviour/use-cases is really best explained using examples, but I believe those examples are better placed in a "learning to script" or "scripting best practices" area elsewhere (which I don't believe yet exists). That'll be an exercise for another day however...

pkubanek (approved):
Looks good.

python/lsst/ts/salobj/topics/read_topic.py
r-owen: I thought about this some more and decided to deprecate the capability for get to flush the queue. Unfortunately this has to be done in stages (to avoid breaking code), and the first stage is to issue a deprecation warning if flush=True, or if the flush argument is omitted, since True is the default. I implemented that here.

If we want to be fully careful then there are 3 more stages (for later versions of salobj):
* Raise if flush=True (or the argument is omitted). We can skip this step if we can prove that all code has been updated to use flush=False.
* Make the default False and warn if the argument is explicitly specified. Raise if the flush=True.
* Eliminate the argument.

r-owen:
Some unit tests are failing because Jenkins using the wrong version of ts_sal. I have pushed a temporary ticket branch to ts_sal and restarted the tests in hopes that will fix Jenkins.

PR #151, opened 2020-09-29
Fix docs in BaseCscTestCase

r-owen:


python/lsst/ts/salobj/sal_info.py
```diff
@@ -614,6 +614,8 @@ async def close(self):
                 except Exception:
                     pass
             self.domain.remove_salinfo(self)
+        except Exception:
+            self.log.exception(f"close failed")
```
wvreeven, line 618: This is an f-string without argument so please change that.

wvreeven (changes requested):
Please fix the f-string without argument. Otherwise all fine.

wvreeven (approved):
Looks good to me!

PR #159, opened 2020-10-26
DM-27296: update to get DDS QoS settings from ts_ddsconfig instead of ts_idl

r-owen:


tribeiro (changes requested):
Please, don't merge this PR until I have a chance to add the package to the develop-env container or it will break everything!

couger01 (approved):
LGTM.

conda/meta.yaml
```diff
@@ -36,6 +36,7 @@ test:
     - boto3
```
tribeiro, line 36: can you update the test requirements to use the new `ts-conda-build` meta package? should be something like:

```
    - ts-conda-build
    - ts-ddsconfig
    - ts-dds
    - ts-idl
```

tribeiro (approved):

PR #160, opened 2020-10-28
DM-26497: add --state and --settings arguments to BaseCsc.amain

r-owen:


python/lsst/ts/salobj/base_csc.py
```diff
@@ -148,7 +180,37 @@ async def start(self):
...
+            # current state to initial_state.
+            # Use the current state instead of default_initial_state
+            # because of Hexapod and Rotator, which do not know their
+            # current state until they connect to the low-level controller.
+            command_state_list = state_transition_dict[
+                (self.summary_state, self._initial_state)
+            ]
+            state_transition_commands = [item[0] for item in command_state_list]
+            try:
+                self._enable_summary_state_event = False
```
mareuter, line 201: I don't like the idea of suppressing the summary state events for automatic transitions. If I look at EFD records I should find no difference between an automatic state changer and a commanded CSC. Otherwise, it looks like the automatic one magically got to that state.

mareuter (changes requested):
Looks pretty good except for the summary state event suppression. See comment.

python/lsst/ts/salobj/base_csc.py
r-owen: I agree it makes me a bit nervous as well. But I think we need it, or a new generic event, instead.

The idea is that when the user sees a summaryState even then they know that the CSC will respond to SAL commands. Without such a rule a user has no idea when the CSC has finished starting up and will listen to SAL commands.

This has always been an issue, but it is made much worse by the fact that the CSC may go through several states before it is ready to be commanded.

It also reduces unit test breakage because we don't get extra summary state messages. (I could probably modify BaseCscTestCase to ameliorate the problem).

r-owen: As agreed at a "standup" meeting, I modified the code as follows:
* Do NOT suppress summaryState events (addressing your concern).
* Enable command callbacks as soon as the first summaryState is output (before that the CSC is likely to be only partially running and commands may fail for confusing reasons).
* Update BaseCscTestCase.make_csc to check all except the final summary state. This will eliminate code breakage in tests that use that method to create CSCs (which should be most of them).

mareuter (approved):
Thank you for talking the time to make the adjustments to keep the summary states in the CSC transitions. Other than one minor comment, this looks good to me.

PR #161, opened 2020-10-30
DM-27380: fix exceptions in SalLogHandler

r-owen:


mareuter (approved):
Looks good to me. Nice detective work.

PR #164, opened 2020-11-16
Fix the conda recipe and use the pre-commit package.

r-owen:


tribeiro (approved):
Thanks for the updates to the conda recipe! looks good.

PR #166, opened 2020-11-30
Add conda install instructions

couger01:


tribeiro:
@couger01, you need to rebase this branch.

doc/index.rst
```diff
@@ -25,6 +25,24 @@ Using lsst.ts.salobj
     sal_scripts
     :maxdepth: 2
 
+
+Conda Install
+-------------
+ts_salobj is offered as a conda package.
```
tribeiro, line 31: Note that we also offer salobj in pypi, so you can actually do `pip install ts-salobj` as well.

tribeiro (changes requested):
This is a good start but I think it would be good to expand this with all the information needed to install salobj on centos. From installing OpenSplice with yum, all the way to having salobj up and running, both with conda and pypi.

pkubanek (approved):

pkubanek (approved):
I wrote something on similar tone for installing, but never merged it. Thanks for providing this missing piece.

PR #171, opened 2021-01-05
DM-28211: CscCommander: round telemetry values to 2 digits, instead of 4, to reduce noise.

r-owen:


tribeiro (approved):
This looks fine, though I wonder if it would be best/feasible to make it a parameter of something like that. But maybe users would deal with it when subclassing?

r-owen:
I agree it is not elegant. A single constant is too crude (though perhaps sufficient for a class that I was hoping would not be much needed). A dict of name: digits seems excessive, but something like that would be the most flexible. Subclasses can handle this in at least two ways: the most flexible is to override callbacks on a topic-by-topic basis. The crudest is to override one of these methods and pick a different number of digits.

PR #178, opened 2021-02-16
DM-28847: Remove test_no_commands from test_sal_info.py

r-owen:
Because in ts_xml 8 we no longer have a SAL component with no commands


r-owen:
Docs are failing to build. I have asked @jonathansick for advice. I hope to fix this before merging, but don't want to hold up ts_xml 8

r-owen:
I used @jonathansick 's advice to update doc/conf.py and the docs now build. Unfortunately Jenkins can't upload them: it complains that `ltd` is not available. I think this used to work. I decided to merge and release anyway as I didn't want to hold up release of ts_xml 8 and this is likely a problem in the Docker image we use.

PR #180, opened 2021-02-19
DM-28892: BaseCsc: improve --simulate command-line argument help and simulation_mode constructor argument handling

r-owen:


python/lsst/ts/salobj/config_schema.py
```diff
@@ -0,0 +1,64 @@
+# This file is part of ts_salobj.
+#
+# Developed for the LSST Telescope and Site Systems.
+# This product includes software developed by the LSST Project
```
wvreeven, line 4: Please use Vera Rubin Observatory at least in line 3 but preferrably also in line 4.

wvreeven (approved):
One minor comment but otherwise this looks good to me.

python/lsst/ts/salobj/config_schema.py
r-owen: I updated line 3. I'm leaving line 4 for now as I don't know what else to say (and it matches all my other packages).

wvreeven: Great. Yes we do not really have a consensus for line 4 AFAIK.

r-owen: I also filed https://jira.lsstcorp.org/browse/DM-28930 in hopes of getting a more official answer (though likely too late for this ticket)

PR #186, opened 2021-03-31
DM-29545: BaseCsc: improve error handling in the constructor

r-owen:


python/lsst/ts/salobj/base_csc.py
```diff
@@ -183,7 +177,15 @@ def __init__(
                     f"simulation_mode={simulation_mode} "
                     f"not in valid_simulation_modes={self.valid_simulation_modes}"
                 )
-        self.evt_simulationMode.set(mode=int(simulation_mode))
+
+        self._requested_simulation_mode = int(simulation_mode)
```
mareuter, line 181: If this variable is gone, why is this reference here? Is this only used as a safety operator? Found another reference in the start function.

python/lsst/ts/salobj/base_csc.py
```diff
@@ -167,12 +167,6 @@ def __init__(
...
 
-        # Postpone assigning command callbacks until `start` is done (but call
-        # assert_do_methods_present to fail early if there is a problem).
-        super().__init__(name=name, index=index, do_callbacks=True)
-
-        # Handle simulation_mode after constructing the base class,
-        # so that evt_simulationMode is available.
         if self.valid_simulation_modes is None:
             warnings.warn(
                 "valid_simulation_modes=None is deprecated", DeprecationWarning
```
mareuter, line 172: If this being set to None is deprecated, the attribute docs should be updated.

python/lsst/ts/salobj/base_csc.py
```diff
@@ -183,7 +177,15 @@ def __init__(
...
+        self._requested_simulation_mode = int(simulation_mode)
+        self._settings_to_apply = settings_to_apply
+        self._summary_state = State(self.default_initial_state)
+        self._initial_state = initial_state
+        self._faulting = False
+        self._heartbeat_task = base.make_done_future()
+        # Interval between heartbeat events (sec)
+        self.heartbeat_interval = HEARTBEAT_INTERVAL
 
         if not hasattr(self, "version"):
```
mareuter, line 190: I noticed that this class variable is not documented in the class attributes section.

mareuter (commented):
A couple of small doc items found. Can you clarify the situation with self._requested_simulation_mode? Seems to still be present in the code.

python/lsst/ts/salobj/base_csc.py
r-owen: Good catch. Added.

python/lsst/ts/salobj/base_csc.py
r-owen: Good suggestion. Done.

python/lsst/ts/salobj/base_csc.py
r-owen: Good catch. I've removed them. A global search showed no use in other packages (and it is a private variable).

mareuter (approved):
Thanks for all the updates. Looks good to me now.

PR #192, opened 2021-05-14
DM-30215: Remove # comment handling and allow quoted parameters.

ktlim:


r-owen (approved):
Please update docs/version_history.rst before merging. I suggest incrementing the 3rd field to make the new version number, since this is minor and fully backwards compatible.

PR #193, opened 2021-05-17
configuration.rst: remove duplicated env var docs

r-owen:
More environment variables are documented in ts_ddsconfig (DM-29713)
so remove the remaining duplication.


mareuter (approved):
Looks fine to me.

PR #194, opened 2021-06-08
DM-30661: make tests/test_salobj_to_either.py compatible with ts_sal 6.

r-owen:


python/lsst/ts/salobj/validator.py
```diff
@@ -36,6 +36,30 @@ class DefaultingValidator:
...
+                default: 2
+              subdict2:
+                type: object
+                properties:
+                  number3:
+                    type: number
+                    # default is ignored this deep; set it at a higher level
+                default:
+                  number3: 3
+
```
wvreeven, line 62: Thanks for adding this. That's very useful.

The issue I was struggling with is that skipping a default on one level, makes that level and all sub-levels completely get removed when the configuration is loaded. Here is what I tried. Note that the `devices` property does not have a default value.

properties:
  devices:
    type: array
    default:
    items:
      type: object
      properties:
        name:
          description: Name of the sensor
          type: string
          default: EssTemperature4Ch
        channels:
          description: Number of channels
          type: integer
          default: 4
        type:
          description: Type of the device
          type: string
          enum:
          - FTDI
          - Serial
          default: FTDI

Sorry for the formatting. GitHub doesn't allow for extensive formatting in comments. Maybe we should discuss this on Slack?

wvreeven (approved):
Looks good to me. Very useful addition to the DefaultingValidator documentation as well.

python/lsst/ts/salobj/validator.py
r-owen: I commented on the Jira ticket, but I think that question may be getting a bit too detailed.

PR #195, opened 2021-06-12
DM-30082: add utc_from_tai_unix

r-owen:


afausti (approved):
I have verified that the new `utc_from_tai_unix` function correctly creates a UTC timestamp from a TAI timestamp running integration tests with `ts_salkafka`, Kafka, and InfluxDB.

PR #196, opened 2021-06-18
DM-30739: Fixed the versions of astropy and numpy.

wvreeven:


pkubanek:
Is there any reason to not make it >=? Just a single version might be too restrictive..

wvreeven:
Yes there is. With `>=` astropy 4.2 gets chosen which introduces all kinds of issues with the conda package dependencies that we are solving by using only `=`.

mareuter (approved):
This is what we want. Good to merge.

PR #197, opened 2021-06-23
DM-30859: update test_idl_parser to allow the (deprecate) private_host field to be absent

r-owen:
Note that whether it is present depends on the version of ts_sal used to create the IDL files.
As such, that field must either be present for all topics or absent for all topics.


tribeiro (approved):
Looks good.

PR #198, opened 2021-06-24
DM-30875: Make NaN compare equal in DDS data

r-owen:
in WriteTopic.set, when deciding if values have changed
and in TestCsc.assert_x_equal methods.


wvreeven (approved):
Thanks for the informative discussion on Slack and the code changes following that. Looks good to me.

PR #200, opened 2021-07-06
DM-31000: hide version.py from mypy

r-owen:


wvreeven (approved):
Looks good to me. I will add this to the confluence page for reference.

PR #202, opened 2021-08-02
DM-31286: update set_random_lsst_dds_partition_prefix to not use "." in the name

r-owen:


pkubanek (approved):
Replaces . in (random) partition prefix with _, tests.

tribeiro:
Here is a small suggestion for an alternative to setting the partition prefix. 

You could convert the the string `{hostname}-{curr_time}-{random_int}` into a hash code. e.g.:

```
import hashlib

hashlib.sha1(f"{hostname}-{curr_time}-{random_int}".encode()).hexdigest()
```

This generates a string that is 40 characters long (2 more than the original). You could probably get only the first 16 characters...

```
hashlib.sha1(f"{hostname}-{curr_time}-{random_int}".encode()).hexdigest()[:16]
```

PR #204, opened 2021-08-19
Tickets/dm 31468

tribeiro:


tests/test_speed.py
```diff
@@ -136,7 +146,7 @@ async def make_remote_and_topic_writer(
                 await asyncio.wait_for(process.wait(), timeout=STD_TIMEOUT)
         finally:
             if process.returncode is None:
-                print("Warning: killing the topic writer")
+                self.log.debug("Warning: killing the topic writer")
```
r-owen, line 149: I would *much* rather that you leave the print statements alone. When I run this test with `pytest -v tests/test_speed.py -s` to see the data locally I don't want to have to specify a hard-to-remember flag such as `--log-cli-level 20` (both the name and the level are non-obvious, whereas I have `-s` hard-wired).

The *point* of this test is to *print results* (and, if possible, publish them). So print statements are arguably correct here.

I tried mixing log messages and print statements and the results seemed fine to me. They are output in the right order, which is the most important thing to my mind. I realize you are not as happy with it, but I'm not sure why.

Note: I could also live with log messages that are reliably output. But you objected to making `--log-cli-level 20` the default (in setup.cfg). That's fine with me. I prefer print statements in unit tests anyway, because they simpler, less work, and I *know* they will be output.

tests/test_speed.py
```diff
@@ -155,15 +165,14 @@ async def test_class_creation_speed(self) -> None:
             dt = time.monotonic() - t0
             ntopics = len(topic_names)
             creation_speed = ntopics / dt
-            print(
+            self.log.debug(
```
r-owen, line 168: This message is important data. It must be info level. That is true of all the other messages that contain measured results.

r-owen (changes requested):
Please restore the print statements in test_speed.py. They serve a very important purpose. They need to be *reliably* output and log messages aren't because the default is to show those only if something fails, and then to only show warnings and errors.

I think of them as as messages at a level between warning and info, perhaps "important info". It is simpler to print than use a logger, but if you want to use a logger than please find a way to make those former print statements reliably output, preferably without enabling all info-level log messages and without them appearing as warnings.

r-owen (approved):
Looks great.

PR #205, opened 2021-09-03
DM-31616: add dependency on ts_utils and deprecated wrappers for moved code

r-owen:


doc/version_history.rst
```diff
@@ -6,6 +6,29 @@
...
+
+Changes:
+
+* Moved basic functions to ts_utils, to make them available with fewer dependencies:
+  
+  * ``current_tai`` and similar time functions.
+  * ``angle_wrap_center`` and similar angle functions.
+  * ``make_done_future``.
+  * test utilities ``assertAnglesAlmostEqual`` (called ``assert_angles_almost_equal`` in ts_utils) and ``modify_environ``.
+* Added temporary wrappers for the code that was moved, for backwards compatibility.
```
wvreeven, line 20: I could be wrong but isn't an empty line above this one required?

wvreeven (approved):
Looks good to me.

doc/version_history.rst
r-owen: It's not needed. It's true that there has to be a blank line before each list begins, and also a blank line before each deeper level of a hierarchical list begins, but not at the transition from a deeper level to a shallower level.

PR #206, opened 2021-09-13
DM-31744: eliminate some deprecation warnings by using ts_utils functions in library code instead of the deprecated wrappers

r-owen:


tribeiro (commented):
Found a couple more:

`current_tai`:

https://github.com/lsst-ts/ts_salobj/blob/ebd049b2832a53a686546d692b2666c39a6fe630/python/lsst/ts/salobj/topics/write_topic.py#L200

https://github.com/lsst-ts/ts_salobj/blob/ebd049b2832a53a686546d692b2666c39a6fe630/python/lsst/ts/salobj/base_script.py#L316

python/lsst/ts/salobj/base.py
```diff
@@ -194,6 +194,9 @@ def astropy_time_from_tai_unix(tai_unix: float) -> astropy.time.Time:
         TAI time as unix seconds, e.g. the time returned by CLOCK_TAI
         on linux systems.
     """
+    warnings.warn(
+        "Use lsst.ts.utils.astropy_time_from_tai_unix instead", DeprecationWarning
+    )
     tai_mjd = (MJD_MINUS_UNIX_SECONDS + tai_unix) / SECONDS_PER_DAY
```
tribeiro, line 200: Should you call the method from `utils` as well?

```
return utils.astropy_time_from_tai_unix(tais_unix=tais_unix)
```

tests/test_basics.py
```diff
@@ -158,12 +143,11 @@ def test_astropy_time_from_tai_unix(self) -> None:
...
             with self.subTest(dt=dt):
                 utc_unix = unix_time0 + dt
-                tai_unix = salobj.tai_from_utc(utc_unix)
-                astropy_time1 = salobj.astropy_time_from_tai_unix(tai_unix)
-                self.assertIsInstance(astropy_time1, astropy.time.Time)
-                self.assertEqual(astropy_time1.scale, "tai")
-                tai_unix_round_trip1 = salobj.tai_from_utc(astropy_time1)
-                self.assertAlmostEqual(tai_unix, tai_unix_round_trip1, delta=1e-6)
+                tai_unix = utils.tai_from_utc(utc_unix)
+                with self.assertWarns(DeprecationWarning):
```
tribeiro, line 147: Maybe add `# TODO: (DM-31660) Remove this deprecated wrapper`.

I also noticed that the other comments in `bade.py` do not mark it as `TODO`. No big deal but you may want to add "TODO".

tribeiro (approved):
Looks good.. I left a few comments inline. I noticed that `astropy_time_from_tai_unix` is still using inline call instead of calling the new method in `utils`. The other methods are calling from `utils` so maybe update it as well?

I also noticed that the comments in the are not using the "TODO" tag. You may want to update that and also add "TODO" comments to remove the deprecation warning checks in the tests. But I think most of those tests are going to be removed, so, up to you.

tests/test_basics.py
r-owen: I changed them all to `# TODO DM-31660:

r-owen:
Good catch on `astropy_time_from_tai_unix`. I have updated it to be a proper wrapper function.

python/lsst/ts/salobj/base.py
r-owen: Good catch. Fixed. I also added a # TODO comment just before the function.

PR #208, opened 2021-09-20
DM-31842: BaseCsc.start: add a brief delay to initial state transitions

r-owen:


mareuter (approved):
Looks good to me. Thanks for the change.

PR #213, opened 2021-11-02
DM-32433: use ADLink's new ddsutil.py instead of a locally cached copy

r-owen:


pkubanek:
Can we expect any improvements or was _ddsutil 1:1 copy taken from Open splice? I am asking as I am experiencing (yet to be resolved) problems with getting history state from Sal/DDS, a d this might fix those

r-owen:
The new official ddsutil.py is identical to our old copy (except adding my two new functions for creating DDS topics). I think problems with historical data are most likely to be from the dds library or QoS settings, not the high-level functions in ddsutil.py

tribeiro:
We need to wait until we roll the library out before merging this PR. Please, don't force merge it or it will break everything!

setup.cfg
```diff
@@ -12,7 +12,7 @@ exclude =
   _ddsutil.py
 
 [tool:pytest]
-addopts = --flake8 --mypy --ignore-glob=*version.py --ignore-glob=*_ddsutil.py
+addopts = --black --flake8 --mypy --ignore-glob=*/version.py --ignore-glob=*/_ddsutil.py
```
tribeiro, line 15: Do you still need `--ignore-glob=*/_ddsutil.py`?

tribeiro (approved):
Looks good! thanks for the update and patience waiting the dds library to be available!

PR #216, opened 2021-12-17
Updated references to point to the main branch.

rbovill:


tests/data/contained_object_schema.yaml
```diff
@@ -1,5 +1,5 @@
 $schema: http://json-schema.org/draft-07/schema#
-$id: https://github.com/lsst-ts/ts_scheduler/blob/master/schema/Scheduler.yaml
+$id: https://github.com/lsst-ts/ts_scheduler/blob/main/schema/Scheduler.yaml
```
r-owen, line 2: You found an existing error. Please change this to:
```
$id: https://github.com/lsst-ts/ts_salobj/blob/master/tests/data/contained_object_schema.yaml
```

r-owen (approved):
Thanks for doing this. One requested update to fix an existing error.

tests/data/contained_object_schema.yaml
rbovill: fixed

PR #217, opened 2021-12-20
DM-32995: output errorCode(errorCode=0) when not in fault state.

r-owen:
And log a critical level message when going to fault state.


rbovill (approved):
LGtM

PR #219, opened 2022-01-06
DM-33152: fix test_dds_get_version to handle dds has `__version__` attribute

r-owen:


tribeiro (approved):
Looks good! thanks for the update

PR #220, opened 2022-01-11
DM-33207: improve reading historical data for an indexed SAL component read with index=0 (meaning "all instances")

r-owen:


pkubanek (approved):
Looks good.

PR #227, opened 2022-03-01
Fix test_sal_info for deleted generic topics

r-owen:
and make the test pickier now that CSCs only get
the commands that they explicitly ask for.


tribeiro (approved):
Looks good.

PR #228, opened 2022-03-04
DM-33941: SalLogHandler.emit: keep track of unfinished tasks

r-owen:
and await them in Controller.close


wvreeven (approved):
Thanks for this change.

r-owen:
I tried a different approach which seems work in the case you gave me. The problem seems to have been caused by Salinfo trying to log messages from within the DDS blocking read thread, which has no event loop. So I moved the queue length checker out of the blocking read.

wvreeven (approved):
Thanks. This indeed fixes the issue for me.

PR #229, opened 2022-03-14
DM-34038: ConfigurableCsc: fix config overrides for hierarchical schema

r-owen:
Add a hierarchical_update function and use it in ConfigurableCsc.
Also work around a few mypy complaints when using the latest mypy
and tweak formatting in one bit of code to be compatible with black 20.8b1 and 22.1.0.


wvreeven (approved):
Looks good to me. Thanks for these changes.

PR #230, opened 2022-03-16
DM-17157: Stop ignoring optional extra commands

r-owen:


r-owen:
Agreed. I tested ts_scriptqueue, ts_mtmount, ts_mthexapod, and ts_mtrotator (the latter two because they used to have extra commands specified in SALSubsystems.xml AddedGenerics).

PR #233, opened 2022-04-08
DM-34384: BaseCsc: make start easier to use. SalLogHandler: allow logging from threads

r-owen:


r-owen:
Thanks. My local mypy didn't catch that; I guess it's too old. I guess record.process can be None, though that seems unlikely once it's been formatted (and the older mypy didn't report that). I've updated the code to work around it.

PR #234, opened 2022-04-13
DM-34440: ConfigurableCsc: always publish evt_configurationApplied when going from standby to disabled

r-owen:


wvreeven (approved):
Looks good to me.

PR #237, opened 2022-04-20
DM-34479: replace descr argument with kwargs in BaseScript.amain and make_from_cmd_line class methods

r-owen:


python/lsst/ts/salobj/base_script.py
```diff
@@ -214,49 +208,72 @@ def make_from_cmd_line(
...
+            if the command specifies ``--schema``.
 
         Notes
         -----
         The final return code will be:
 
         * 0 if final state is `lsst.ts.idl.enums.Script.ScriptState.DONE`
-          or `lsst.ts.idl.enums.Script.ScriptState.STOPPED`
+          or `lsst.ts.idl.enums.Script.ScriptState.STOPPED`.
         * 1 if final state is `lsst.ts.idl.enums.Script.ScriptState.FAILED`
```
wvreeven, line 238: I see that you added a terminating `.` to line 237 so maybe add one here too?

python/lsst/ts/salobj/base_script.py
```diff
@@ -214,49 +208,72 @@ def make_from_cmd_line(
...
-        * 2 otherwise (which should never happen)
+        * 2 if the final script state is other than DONE, STOPPED, or FAILED,
+            or if the final script state is DONE but script cleanup raises.
+            This should never happen
```
wvreeven, line 241: I see that you added a terminating `.` to line 237 so maybe add one here too?

python/lsst/ts/salobj/base_script.py
```diff
@@ -214,49 +208,72 @@ def make_from_cmd_line(
...
+            Keyword arguments for the script constructor.
+            Must not include ``index``.
+            Ignored (other than testing for ``index``)
+            if the command specifies ``--schema``.
 
         Notes
         -----
         The final return code will be:
 
         * 0 if final state is `lsst.ts.idl.enums.Script.ScriptState.DONE`
```
wvreeven, line 236: These return codes look like ideal candidates to extract into an enum. Maybe for a future ticket?

wvreeven (approved):
Looks good to me. I left a few comments for improvements and leave it to you to process or not.

python/lsst/ts/salobj/base_script.py
r-owen: I am actually leaning towards removing all these special return codes and simply returning 1. The ScriptQueue is not using them and I can't think of a use for them. The script reports its final state using SAL.

r-owen: I decided to simplify exit codes to 0/1 after looking at ts_scriptqueue and finding no usage of the exit code. All unit tests pass in ts_scriptqueue.

PR #239, opened 2022-05-04
DM-34661: Controller and SalInfo: add write_only constructor arg

r-owen:


python/lsst/ts/salobj/controller.py
```diff
@@ -84,6 +84,11 @@ class Controller:
         Set ``do_<name>`` methods as callbacks for commands?
         If True then there must be exactly one ``do_<name>`` method
```
wvreeven, line 85: In lines 85 to 91 (and possibly more, I didn't check), you're mixing `True` with `true`. Please pick one and stick to that.

EDIT: line 123 also contains `true`.

wvreeven (approved):
Looks good to me. I have one comment about mixing `True` with `true` and leave it for your consideration.

PR #242, opened 2022-05-18
DM-34837: Clarify WriteTopic set_write docs

r-owen:


python/lsst/ts/salobj/topics/write_topic.py
```diff
@@ -270,25 +270,23 @@ def set(self, **kwargs: typing.Any) -> bool:
     async def set_write(
         self, *, force_output: typing.Optional[bool] = None, **kwargs: typing.Any
     ) -> SetWriteResult:
-        """Set zero or more fields of ``self.data`` and write if changed
-        or if ``force_output`` true.
+        """Set zero or more fields of ``self.data`` and write the data if any
+        of the specified fields have changed or if ``force_output`` is true.
```
pkubanek, line 275: Even better?

of the specified fields have changed, it was the first field call to set or set_write or if ``force_output`` is true.

pkubanek (approved):
I like the changes. See small clarification regarding first set/set_write calls, which I found important.

python/lsst/ts/salobj/topics/write_topic.py
r-owen: I'm afraid I don't understand this text. Also I am very reluctant to `set` in the brief description for `set_write` as I am afraid of giving the false impression of a tighter interaction between those calls then actually exists. I have moved all the information about whether `set_write` writes the data to the description of the `force_output` parameter and tried to make that as clear as I could.

I have taken another pass at cleaning up the docs for set_write, set, and has_data. Please have another look to see if it is now sufficiently clear. `set` already had a section on why it exists and how to use it with `set_write` but I expanded that. The documentation for `has_data` was too brief. I hope it is clearer now.

pkubanek (approved):
I like has_data updates, but others are also great.

PR #243, opened 2022-05-25
U/rowen/logauthorizeonoff: log LSST_DDS_ENABLE_AUTHLIST at INFO instead of DEBUG, and update the version history

r-owen:


r-owen:
This update includes the new info that the next ts_xml will be 11.2 instead of 12, and the next ts_sal will be 7 instead of 6.2

wvreeven (approved):
Looks good to me.

PR #244, opened 2022-05-30
Tickets/dm 34702

wvreeven:


conda/meta.yaml
```diff
@@ -32,9 +35,10 @@ test:
     - tests
     - schema
     - setup.cfg
+    - pyproject.toml
     - .git
-  commands:
-    - pytest
+#  commands:
+#    - pytest
```
r-owen, line 41: Please remove these comment-out lines, if practical. If you want to keep them then please add a comment explaining why they are commented-out and when they can be uncommented (or discarded), preferably with a JIRA ticket #.

python/lsst/ts/salobj/cli.py
```diff
@@ -0,0 +1,35 @@
+# This file is part of ts_salobj.
```
r-owen, line 1: Please delete this file. It is no longer being used.

r-owen (approved):
Looks great. I had a few small suggestions.

conda/meta.yaml
wvreeven: Thanks for catching that. Those should not be uncommented. I just did it to be able to run the build quicker. I will uncomment them again.

python/lsst/ts/salobj/cli.py
wvreeven: Done

PR #245, opened 2022-06-08
Tickets/dm 35131

wvreeven:


conda/meta.yaml
```diff
@@ -53,7 +53,7 @@ requirements:
     - python
     - setuptools
     - setuptools_scm
-    - numpy =1.20
+    - numpy >=1.20
```
r-owen, line 56: Consider ~=1.20. I think that means "any version that should be compatible with 1.20"

r-owen (approved):
Thanks for fixing this! One change to consider.

PR #249, opened 2022-08-24
DM-36001: Make sure that all received data are processed

pkubanek:


python/lsst/ts/salobj/sal_info.py
```diff
@@ -781,7 +785,8 @@ def _blocking_read(
         conditions = self._waitset.wait(self._wait_timeout)
         if not self.isopen:
             # shutting down; clean everything up
-            return (None, [])
+            return None
+        ret = []
```
r-owen, line 781: This needs a type annotation to make mypy happy

python/lsst/ts/salobj/sal_info.py
```diff
@@ -792,8 +797,9 @@ def _blocking_read(
             data_list = [
                 self._sample_to_data(sd, si) for sd, si in data_list if si.valid_data
             ]
-            return (reader, data_list)
-        return (None, [])
+            ret.append((reader, data_list))
```
r-owen, line 792: Please only append if data_list is not empty, e.g.:
```
if data_list:
    ret.append((reader, data_list))
```
(I probably should have done that with the old code, though it fits more neatly into the new code).

python/lsst/ts/salobj/sal_info.py
```diff
@@ -781,7 +785,8 @@ def _blocking_read(
         conditions = self._waitset.wait(self._wait_timeout)
         if not self.isopen:
             # shutting down; clean everything up
-            return (None, [])
+            return None
```
r-owen: Please return an empty list, to simplify the API.

python/lsst/ts/salobj/sal_info.py
```diff
@@ -766,12 +765,17 @@ async def _read_loop(self, loop: asyncio.AbstractEventLoop) -> None:
 
     def _blocking_read(
         self,
-    ) -> typing.Tuple[
-        typing.Optional[topics.ReadTopic], typing.List[type_hints.BaseMsgType]
+    ) -> typing.Optional[
```
r-owen: Please remove typing.Optional and return an empty list if nothing is read.

Also please consider changing `typing.List` to `list` and `typing.Tuple` to `tuple`. I plan to take a global pass on salobj for this, but meanwhile you may as well update new code?

python/lsst/ts/salobj/sal_info.py
r-owen: Please update the doc string to something along these lines:
```
"""Read DDS data

Returns
-------
read_data
    Read data, as a list of tuples, each with two values:

    * reader : `ReadTopic`
        The read topic.
    * data_list: `list` [`BaseMsgType`]
        List of one or more messages.
"""
```
The most important changes are:
* reader cannot be None, so simplify that description.
* data_list cannot be empty (if you update the code below as requested).
* Move the return description to a Returns section, fixing an existing issue.

python/lsst/ts/salobj/sal_info.py
```diff
@@ -747,15 +747,14 @@ async def _read_loop(self, loop: asyncio.AbstractEventLoop) -> None:
...
+                    reads = await loop.run_in_executor(pool, self._blocking_read)
                     if not self.isopen:
                         break
-                    if not data_list or reader is None:
-                        continue
-                    reader.dds_queue_length_checker.check_nitems(len(data_list))
-                    reader._queue_data(data_list)
+                    for reader, data_list in reads:
+                        if not data_list:
+                            continue
```
r-owen, line 755: Please delete this `if not data_list:` block. With a suggested change below the list will never be empty. And even if something does go wrong, the code will still be safe -- this is only intended as a no-longer-needed small optimization.

python/lsst/ts/salobj/sal_info.py
```diff
@@ -766,22 +765,20 @@ async def _read_loop(self, loop: asyncio.AbstractEventLoop) -> None:
...
         self,
-    ) -> typing.Tuple[
-        typing.Optional[topics.ReadTopic], typing.List[type_hints.BaseMsgType]
+    ) -> typing.List[
+        typing.Tuple[topics.ReadTopic, typing.List[type_hints.BaseDdsDataType]]
     ]:
         """Read DDS data.
 
-        Return two values:
+        Return array of tuples (two values):
```
r-owen, line 773: Please update the doc string with new info. Assuming you return an empty list instead of None and assuming you don't return empty data_lists, something like:
```
"""Read DDS data

Returns
-------
read_data
    Read data, as a list of tuples, each with two values:

    * reader : `ReadTopic`
        The read topic.
    * data_list: `list` [`BaseMsgType`]
        List of messages.
"""
```

r-owen (changes requested):
This looks like a very useful update. I have a few small requested changes, and the larger request to add a new entry to the version history (bumping the smallest version by one).

python/lsst/ts/salobj/sal_info.py
pkubanek: Removed Optimal

pkubanek:
I incorporated your suggestions, added v7.1.3 entry. Please note that most of isopen etc. checks were dropped, as they aren't needed (once reader is available).

python/lsst/ts/salobj/sal_info.py
r-owen: Please make this change (or if you don't want to, please explain why -- you marked this resolved without changing it).

python/lsst/ts/salobj/sal_info.py
r-owen: Please make this change (or if you don't want to, please explain why -- you marked this resolved without changing it).

python/lsst/ts/salobj/sal_info.py
r-owen: Please make this change (you marked this resolved without changing it).

r-owen: Also please put:
```
if not self.isopen:
    return []
```
just after `conditions = ...`, or to be even more cautious, just after `for condition in conditions:`, to avoid unexpected warning messages at shutdown. self.isopen is set false earlier than ReadTopic.isopen, with the event loop running in between, and once self.isopen is false there is no point to continuing.  We can live without it, but I'd rather play it safe for fear of introducing unexpected warnings at shutdown.

r-owen (changes requested):

python/lsst/ts/salobj/sal_info.py
pkubanek: Fixed now, sorry for missing it - I though that was resolved with my previous updates, which wasn't the case

pkubanek:
Updated, please have a look again. Thanks

python/lsst/ts/salobj/sal_info.py
```diff
@@ -792,8 +788,14 @@ def _blocking_read(
...
                 self._sample_to_data(sd, si) for sd, si in data_list if si.valid_data
             ]
-            return (reader, data_list)
-        return (None, [])
+            if not data_list:
+                ret.append((reader, data_list))
+
+        if not self.isopen:
+            # shutting down; clean everything up
+            return []
```
r-owen: The main reason for this test is to avoid trying to read DDS data after the SalInfo is closed, for fear that the DDS system itself may be shut down, which results in confusing messages.

The safest place to put it is in the loop over conditions -- i.e. before reading DDS data from each reader. I moved it there in the commit I just pushed.

It could also go before the loop, which reduces the # of tests if there is more than one condition, but adds a small interval over which the system may be closed but the loop tries to read data. Or we could skip it, but I'd prefer to test it at least once before trying to read any data from a DDS reader.

r-owen (approved):
Please look at the change I just pushed, which I think fixes a bug that prevents any read data from being returned. If you are OK with it, feel free to squash it or leave it. If you prefer to modify it, go ahead.

python/lsst/ts/salobj/sal_info.py
```diff
@@ -792,8 +788,14 @@ def _blocking_read(
             data_list = [
                 self._sample_to_data(sd, si) for sd, si in data_list if si.valid_data
             ]
-            return (reader, data_list)
-        return (None, [])
+            if not data_list:
```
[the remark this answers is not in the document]
pkubanek: Agree, my fault.

python/lsst/ts/salobj/sal_info.py
pkubanek: I am Ok with that - unaware of Reader interaction, I placed test after the loop. Fine with whenever it ends up.

PR #250, opened 2022-08-31
DM-36072: Fix `ReadTopic.aget` to not take data from `next`

r-owen:


wvreeven (approved):
Looks good to me. I went through the commits one by one since modernizing the type annotations introduced a lot of clutter around the other changes.

PR #255, opened 2022-09-30
DM-36420: fix a potential race condition getting historical data

r-owen:


r-owen:
Note: we don't expect Jenkins to pass because it is not yet set up to run Kafka-based unit tests.

couger01 (approved):
LGTM.

PR #256, opened 2022-10-10
DM-36318: BaseScript: exit if configuration fails

r-owen:


pkubanek (approved):
Changes configuration failure handling as promised.

PR #264, opened 2023-01-07
DM-37501: deprecate synchronous callbacks

r-owen:


python/lsst/ts/salobj/csc_commander.py
```diff
@@ -272,6 +285,20 @@ def __init__(
...
+        all_members = inspect.getmembers(self, predicate=inspect.ismethod)
+        sync_do_member_names = [
+            member[0]
+            for member in all_members
+            if member[0].startswith("do_")
+            and not inspect.iscoroutinefunction(member[1])
+        ]
+        if sync_do_member_names:
+            warnings.warn(
+                f"Synchronous do_ methods are deprecated: {sorted(sync_do_member_names)}",
```
pkubanek, line 298: I have a slight preference for use of ','.join, as listing method names in an array (as [m1,m2,..] is a bit brutal.

python/lsst/ts/salobj/topics/read_topic.py
```diff
@@ -428,6 +430,16 @@ def callback(
...
     @callback.setter
     def callback(self, func: None | CallbackType) -> None:
+        if func is not None:
+            if not callable(func):
+                raise TypeError(f"callback {func} not callable")
+            if not inspect.iscoroutinefunction(func):
+                # TODO DM-37502: modify this to raise (and update doc string)
+                # once we drop support for synchronous callback functions.
+                warnings.warn(
+                    "callback should be asynchronous", category=DeprecationWarning
```
pkubanek, line 440: Can that produce which callback? f"callback {func} should.. perhaps?

pkubanek (approved):
Looks good, see suggestions for minor improvements.

python/lsst/ts/salobj/csc_commander.py
r-owen: I would prefer to keep this simple, and the [] act as a useful delimiter.

PR #271, opened 2023-04-11
DM-38459: add xml enum parsing

r-owen:


dmills-lsst (commented):
Looks ok to me (I only looked at the avro schema related files)

r-owen:
Thank you. I had the wrong base branch so there were a lot more changes than there should have been.

PR #272, opened 2023-05-03
DM-39001: remove scons support; improve async function detection; improve some exception docs

r-owen:


pkubanek (approved):
Nice addition.

PR #279, opened 2023-06-08
DM-37999 kafka: BaseCsc: check if a duplicate CSC is already running, and fail to start if so 

r-owen:


doc/installation.rst
```diff
@@ -13,7 +15,63 @@ or::
...
+
+  * Set the following environment variables::
+
+      export LSST_KAFKA_BROKER_ADDR="127.0.0.1:9092"
+      export LSST_SCHEMA_REGISTRY_URL="http://schema-registry:8081"
+
+* To start and stop Kafka, use the following bash/zsh functions::
+
+    function kafka_run() {
+        pushd /Users/rowen/UW/LSST/tsrepos/ts_salobj
```
tribeiro, line 51: Maybe add a stub here for the path? Something like:

```
        pushd <tsrepos-path>/ts_salobj
```

and say that one must replace `<tsrepos-path>` with the path to their directory with ts_salobj repository?

doc/installation.rst
```diff
@@ -13,7 +15,63 @@ or::
...
+        docker-compose up -d zookeeper broker schema-registry
+        popd
+    }
+
+    function kafka_stop() {
+        pushd /Users/rowen/UW/LSST/tsrepos/ts_salobj
```
tribeiro, line 57: same as above about the path.

tribeiro (approved):
Looks good. Just a small comment on the documentation.

doc/installation.rst
r-owen: Good catch. Thank you!

PR #281, opened 2023-06-22
DM-39753: add MockWriteTopic and make_mock_write_topics

r-owen:


python/lsst/ts/salobj/topics/mock_write_topic.py
```diff
@@ -0,0 +1,91 @@
...
+
+from .. import type_hints
+from .write_topic import MAX_SEQ_NUM, WriteTopic
+
+if typing.TYPE_CHECKING:
+    from ..sal_info import SalInfo
+
+
+class MockWriteTopic(WriteTopic):
+    r"""A version of WriteTopic that records every message written
```
wvreeven, line 36: Maybe I miss something but why is there an `r` in front of the `"""`?

wvreeven (approved):
Looks good to me. I left one minor comment for your consideration.

python/lsst/ts/salobj/topics/mock_write_topic.py
r-owen: I'll remove it. When a string has back-slashes in it, it avoids complaints about unrecognized escape sequences (without the need to double each backslash).

PR #282, opened 2023-08-21
DM-40349: Move schema generation code from salobj-kafka to xml.

tribeiro:


dmills-lsst (commented):
Not passing any test....

dmills-lsst (approved):
OK

PR #289, opened 2023-11-03
DM-40580: Preparation work for kafka deployment at TTS.

tribeiro:
Some additional changes to address event retention as well as some improvements to unit testing support.


wvreeven (approved):
Looks great! My only comment would be to try and squash all commits marked as `fixup` if possible.

PR #295, opened 2024-05-16
Remove the do_setAuthList command.

wvreeven:


tests/test_sal_info.py
```diff
@@ -63,7 +63,7 @@ async def test_salinfo_constructor(self) -> None:
                 with pytest.raises(RuntimeError):
                     salinfo.assert_started()
 
-                asyncio.create_task(salinfo.start())
+                await asyncio.create_task(salinfo.start())
```
tribeiro, line 66: Is this change right? I think the idea was to run it in the background (hence, no `await`) and than verify that the `start_task` is marked as done at the end.

tribeiro (approved):
Looks  good... I did left one inline comment I think you should consider before merging it. 

I think adding the `await` in the test there is actually against what the test is doing.

Also, a small pet peeve, I noticed you did more than just remove the `do_setAuthList` command. There some small other things sprinkled here and there. I would really recommend you split these changes into different commits. the way it is right now you have all the changes in a single commit, which can be misleading.

tests/test_sal_info.py
wvreeven: Thanks for pointing this out. I reverted the change.

PR #297, opened 2024-08-19
fix PySide6 compatibility

pkubanek:
Due to incompatibilities of different AsyncIO implementation, it's
better to call run_in_loop without specifying func argument by name.


tribeiro (approved):
Looks good!

PR #298, opened 2024-08-21
tickets/DM-45216

iglesu:
This PR introduces a new `WildcardIndexError` exception to handle wildcard indices (`*`) in SAL component names. This is particularly useful in cases where an index in a component name (e.g., `Test:*`) is a wildcard, allowing operations to be applied to all indices of a given CSC.

Key changes:
- Added `WildcardIndexError` to handle wildcard indices in SAL components, ensuring backward compatibility.
- Applied this in `ts_standardscripts`'s `set_summary_state.py` to support valid wildcard cases.
- Added unit tests in `name_to_name_index` to ensure:
  - Backward compatibility by raising `ValueError` when needed.
  - Correct handling of wildcard indices using `WildcardIndexError`.

### DIFF
--- a/doc/version_history.rst
+++ b/doc/version_history.rst
@@ -9,6 +9,13 @@ Version History
 
 .. Version 8 of salobj will contain the kafka release.
 
+v7.7.1
+------
+
+* Added ``WildcardIndexError`` to handle wildcard indices in SAL components,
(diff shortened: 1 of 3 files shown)


tribeiro (approved):
Looks good!

PR #302, opened 2024-11-27
In sal_info.py, fix issue in initialization that was causing the star…

tribeiro:
…t_task to finish prematurely.


wvreeven (approved):
LGTM

PR #303, opened 2024-11-27
In sal_info.py, remove debug message from high level loop in read_loo…

tribeiro:
…p and add a debug message to indicate when historical data is done for a particular topic.


wvreeven (approved):
LGTM

PR #304, opened 2024-12-13
Update Topics unit test to account for the change in ts-xml that no l…

tribeiro:
…onger enforces arrays size.

This has been converted into a warning instead of an error.

The unit test now checks that it can write and read data with different array lenght successfully, and that it issues a warning when doing so.


mareuter (approved):
Looks good to me.

PR #305, opened 2025-01-08
DM-48231: Fix issue with kafka Consumer group being left behind when closing application.

tribeiro:
Fixing the issue with the consumer groups left behind required updating the confluent-kaka > 2. But we still had that issue with the v2 of the library with the publish delays. After a deep dive into the docs I finally managed to find a solution for it, which allows us now to update to the newest version of the library and also frees us to update to python 3.12.


mareuter (approved):
Looks good to me.

PR #309, opened 2025-02-12
DM-48885: Add support for configuring number of messages to be consumed and timeout in salobj Remote/SalInfo.

tribeiro:


wvreeven (approved):
Looks great!

python/lsst/ts/salobj/sal_info.py
```diff
@@ -1207,126 +1226,137 @@ async def _read_loop(self) -> None:
...
-                    elif (
-                        schema_resolution_errors[kafka_name]
-                        % SCHEMA_RESOLUTION_LOG_ERROR_THRESHOLD
-                        == 0
-                    ):
-                        self.log.error(
-                            f"Failed to deserialize {schema_resolution_errors[kafka_name]} samples of "
-                            f"{kafka_name}. Check schema compatibility!"
+                for message in messages:
+                    message_error = message.error()
```
pkubanek, line 1230: Keeping message processing leads to quite big indentation, which makes code harder to read. Can you please create a method to process messages, so the code (assuming it's called _process_message) will shrink to:
```
for message in messages:
    self._process_messaeg(message)
```

doc/news/README.rst
```diff
@@ -0,0 +1,40 @@
+Recording Changes
+=================
+
```
pkubanek, line 3: There is an empty space. Not a big deal, but if you can get rid of those (:%s/ *$//g in vim), that will save some bits.

PR #318, opened 2025-05-31
tickets/DM-50528

mareuter:


tribeiro (approved):
Looks good... One small comment about the naming though, consider using `check_schema` instead, at least for the executable... 

But up to you...

PR #321, opened 2025-07-28
Sleep some time when deleting the topics in unit tests.

teweitsai:
Sleep some time when deleting the topics in unit tests.

### DIFF
--- a/doc/news/OSW-783.bugfix.rst
+++ b/doc/news/OSW-783.bugfix.rst
@@ -1,0 +1,1 @@
+Sleep some time when deleting the topics in unit tests.
--- a/python/lsst/ts/salobj/base_csc_test_case.py
+++ b/python/lsst/ts/salobj/base_csc_test_case.py
@@ -110,6 +110,9 @@ class BaseCscTestCase(metaclass=abc.ABCMeta):
 
         delete_topics.execute(delete_topics_args)
 
+        # Sleep some time to let the cluster have time to finish the deletion
+        await asyncio.sleep(5.0)
+
     @abc.abstractmethod
     def basic_make_csc(
         self,


teweitsai:
@tribeiro I tested this under my local `docker` environment.

If I run the M2 CSC unit test without this, I will get the uncleared topics (non empty `[]` output when using the `curl http://0.0.0.0:8081/subjects` under the `schema-registry` container).

If I use the above code, the result will be empty (`[]`).

tribeiro:
This is really weird, honestly... can you please hold off while I have a chance to investigate it? in principle this should not be necessary as it is currently waiting for things to be deleted.

teweitsai:
Sure @tribeiro . And you can see my log in OSW-783. Thanks!

tribeiro (approved):
Here is my assessment of the situation:

This is a bit weird because the calls to delete schema is made in 2 stages; we first mark them for deletion [here](https://github.com/lsst-ts/ts_salobj/blob/21c52eb579c5d5d44033e870a96cbc028bacdd32/python/lsst/ts/salobj/delete_topics.py#L249), then permanently deleter them [here](https://github.com/lsst-ts/ts_salobj/blob/21c52eb579c5d5d44033e870a96cbc028bacdd32/python/lsst/ts/salobj/delete_topics.py#L256).

As you can see, those calls are synchronous so, I am not really sure why adding the await makes any difference... 

Digging a bit deeper into the API, I just realized that confluent_kafka is now on version 2.11.0, which introduces substantial changes to how this is done. This means we will have to rework this too to upgrade to newer versions of the library. I think the best course is then to get this merged and we update the code later, hopefully fixing the issue in the process.

PR #336, opened 2026-03-04
Tickets/osw 1886

wvreeven:


tribeiro (approved):
Looks good. The majority of the changes are due to the adoption of ruff, thanks for separating those style changes into its own commit. I mentioned this in the ts-xml ticket but will mention it again here just in case, would be good to add a test for attribute overflow to make sure we are capturing what we intended to with the update.

PR #343, opened 2026-06-10
Tickets/osw 2405

wvreeven:


tribeiro (approved):
Looks good! Pretty neat solution with the set and the callback to clean it up afterwards...

